PR #1, opened 2024-06-25
MM-58804: Initial setup

Rajat-Dabade:
The PR includes:

- Linter fixes
- Updates to package.json scripts
- Dependency fixes
- Updated modd.conf for live watching changes
- Removal of unnecessary files
- Updates to the Makefile workflow

Also, 
Folder `main` duplicates the code for the server and webapp, which was for reference purposes and was pushed by mistake, this PR removes it. 

Regarding documentation i.e. `README`, `CONTRIBUTING.md` etc are in the folder `imp-files`, which will be moved to the root folder in the subsequent PR's.


.github/workflows/dev-release.yml
harshilsharma63, line 1: I don't think we need this file. Delivery team recently updated the plugin release process. Check out details here - https://hub.mattermost.com/private-core/pl/yajxpk58kfne5qakj88apyfsze

harshilsharma63: Same for other release CI files.

main/docs/README.md
harshilsharma63, line 1: Why are we deleting all the docs? Lot of this is plugin/personal server agnostic and is still relevant.

webapp/src/components/messages/versionMessage.tsx
```diff
@@ -88,4 +88,7 @@ const VersionMessage = React.memo(() => {
         </div>
     )
 })
+
+VersionMessage.displayName = 'VersionMessage'
```
harshilsharma63, line 92: I'm not sure what this line is doing.

harshilsharma63 (changes requested):

main/docs/README.md
Rajat-Dabade: We are not deleting the does, it's in the folder `imp-files`, this will be moved to the root in subsequent PR.

webapp/src/components/messages/versionMessage.tsx
Rajat-Dabade: This was added becuase of the linter error, having the displayName property will not create any issue, instead, it will help in debugging and displaying the component names in the error stack trace.

hmhealey: The `displayName` is also set automatically if we define the component as `function MyComponent`, although it's not set if you define it as `const MyComponent = ...`. I'm not sure how to have `React.memo` set it automatically, but maybe it'll do that if you were to write this as `const VersionMessage = React.memo(function VersionMessage(){...});`

webapp/package.json
```diff
@@ -5,11 +5,13 @@
 	"description": "",
 	"scripts": {
 		"build": "webpack --mode=production",
+		"build:watch": "webpack --mode=production --watch",
 		"pack": "cross-env NODE_ENV=production webpack --config webpack.prod.js",
 		"packdev": "cross-env NODE_ENV=dev webpack --config webpack.dev.js",
 		"watchdev": "cross-env NODE_ENV=dev webpack --watch --progress --config webpack.dev.js",
 		"deveditor": "cross-env NODE_ENV=dev webpack server --config webpack.editor.js",
 		"test": "jest",
+		"lint": "eslint --ignore-pattern node_modules --ignore-pattern dist --ext .js --ext .jsx --ext tsx --ext ts . --quiet --cache",
```
hmhealey, line 14: Should this be combined with `check` from below? Calling this `lint` probably makes more sense in general, but since we tend to call this `check` in other modules, it might be good to stay consistent with that

Makefile
```diff
@@ -71,21 +71,6 @@ apply:
...
-ifneq ($(HAS_SERVER),)
-	@if ! [ -x "$$(command -v golangci-lint)" ]; then \
-		echo "golangci-lint is not installed. Please see https://github.com/golangci/golangci-lint#install-golangci-lint for installation instructions."; \
-		exit 1; \
-	fi; \
-
-	@echo Running golangci-lint
-	golangci-lint run ./...
-endif
-
```
hmhealey, line 88: Did you mean to remove this? It looks like `make check-style` doesn't do anything any more.

Also, I think this section is common for all plugin projects, so it might be worth keeping it that way to keep this closer to the other plugins

Makefile
```diff
@@ -423,84 +349,16 @@ server-lint: ## Run linters on server code.
...
-	cd server; go test -tags '$(BUILD_TAGS)' -race -v -coverpkg=./... -coverprofile=server-postgres-profile.coverage -count=1 -timeout=30m ./...
-	cd server; go tool cover -func server-postgres-profile.coverage
-	cd mattermost-plugin/server; go test -tags '$(BUILD_TAGS)' -race -v -coverpkg=./... -coverprofile=plugin-postgres-profile.coverage -count=1 -timeout=30m ./...
-	cd mattermost-plugin/server; go tool cover -func plugin-postgres-profile.coverage
-	docker-compose -f ./docker-testing/docker-compose-postgres.yml down -v --remove-orphans
-
 webapp-ci: ## Webapp CI: linting & testing.
-	cd webapp; npm run check
-	cd mattermost-plugin/webapp; npm run lint
+	cd webapp; npm run check; npm run lint
```
hmhealey, line 360: Does this also run the type checker on the web app code? That's might not be needed with Webpack using `ts-loader` though

hmhealey (approved):
A few comments, but nothing blocking. This was a surprisingly easy 400,000 line PR to review 😛

.github/workflows/dev-release.yml
Rajat-Dabade: Will take care of this in subsequent PR.

webapp/package.json
Rajat-Dabade: Yup, it's a good catch, will combine this with check.

Makefile
Rajat-Dabade: Yup, we can keep this, reverting this back. 🙌

harshilsharma63 (approved):
Approving as comments will be addressed in follow-up PR.

PR #8, opened 2024-07-09
stop logging errors for api error responses

lieut-data:
#### Summary
From a cursory glance, when a real error occurs, we already log something at the application layer, so there's no reason to echo every api error response as an error log too. Errors should be actionable, but these logs were not, e.g.:

> access denied to templates

or:

> category ID specified in input does not exist for user

This PR duplicates https://github.com/mattermost/focalboard/pull/5022.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59325

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -178,7 +178,6 @@ func (a *API) userIsGuest(userID string) (bool, error) {
 // Response helpers
 
 func (a *API) errorResponse(w http.ResponseWriter, r *http.Request, err error) {
-	a.logger.Error(err.Error())
 	errorResponse := model.ErrorResponse{Error: err.Error()}
 
 	switch {
@@ -195,14 +194,15 @@ func (a *API) errorResponse(w http.ResponseWriter, r *http.Request, err error) {
 	case model.IsErrNotImplemented(err):
 		errorResponse.ErrorCode = http.StatusNotImplemented
 	default:
-		a.logger.Error("API ERROR",
-			mlog.Int("code", http.StatusInternalServerError),
-			mlog.Err(err),
-			mlog.String("api", r.URL.Path),
-		)
 		errorResponse.Error = "internal server error"
 		errorResponse.ErrorCode = http.StatusInternalServerError
 	}
+
+	a.logger.Warn("api error response",
+		mlog.Int("code", http.StatusInternalServerError),
+		mlog.Err(err),
+		mlog.String("api", r.URL.Path),
+	)
 
 	setResponseHeader(w, "Content-Type", "application/json")
 	data, err := json.Marshal(errorResponse)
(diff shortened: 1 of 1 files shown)


server/api/api.go
```diff
@@ -195,15 +194,16 @@ func (a *API) errorResponse(w http.ResponseWriter, r *http.Request, err error) {
...
-			mlog.Int("code", http.StatusInternalServerError),
-			mlog.Err(err),
-			mlog.String("api", r.URL.Path),
-		)
 		errorResponse.Error = "internal server error"
 		errorResponse.ErrorCode = http.StatusInternalServerError
 	}
 
+	a.logger.Warn("api error response",
+		mlog.Int("code", http.StatusInternalServerError),
```
marianunez, line 202: Should this be the error code that was set in the cases above instead of the default for all?

```suggestion
		mlog.Int("code", errorResponse.ErrorCode),
```

marianunez:
> From a cursory glance, when a real error occurs, we already log something at the application layer,

@lieut-data can you point to where this is happening? For the two examples provided, I don't see where the error is logged in the application layer. Worried that we may be missing seeing actual errors in the log with changing for all that call `errorResponse`

server/api/api.go
lieut-data: Ah, goodness, yes! Sorry for that obvious slip up :/

lieut-data:
> @lieut-data can you point to where this is happening? For the two examples provided, I don't see where the error is logged in the application layer. Worried that we may be missing seeing actual errors in the log with changing for all that call `errorResponse`

Oops, I didn't meant to frame it as such, @marianunez: the two examples are "non-actionable" errors, and indeed there are no additional error logs expected from the application layer. With this change, we'll still log a warning when an API returns a failure code to the client, but the default position is to assume these aren't actionable (i.e. permission issues).

There are examples of actionable errors in the app layer, e.g.:
https://github.com/mattermost/mattermost-plugin-boards/blob/0b6e4435549392a49e2fc51e49d778a751028737/server/app/initialize.go#L10-L16

but I note they are not consistent. There are certainly opportunities for greater consistency here.

marianunez:
> the two examples are "non-actionable" errors, and indeed there are no additional error logs expected from the application layer. With this change, we'll still log a warning when an API returns a failure code to the client, but the default position is to assume these aren't actionable (i.e. permission issues).

I see, makes sense. Thanks for clarifying

marianunez (approved):

server/api/api.go
```diff
@@ -178,7 +178,6 @@ func (a *API) userIsGuest(userID string) (bool, error) {
 // Response helpers
 
 func (a *API) errorResponse(w http.ResponseWriter, r *http.Request, err error) {
-	a.logger.Error(err.Error())
```
marianunez, line 181: This explains the double log entries for each error

Rajat-Dabade (approved):
LGTM 🚀

PR #15, opened 2024-07-26
Update license to AGPL

marianunez:
#### Summary

License update to release Boards plugin under AGPL license.

#### Ticket Link
None

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,33 @@
-                                 Apache License
+Mattermost Licensing
+
+SOFTWARE LICENSING
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -199,3 +228,670 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
```
marianunez, line 231: @jwilander I copied the same version that Focalboard was using but not sure if this section is required.

alerque: Yes, having a copy of the AGPL text included here is appropriate.

jwilander (approved):

PR #23, opened 2024-10-13
MM-61043 Error deleting board if fileID is empty

Rajat-Dabade:
#### Summary

https://hub.mattermost.com/icu/pl/qqb9tqrzdtfajyfrysd39ecaro

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61043

### DIFF
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -68,7 +68,7 @@ func (a *App) GetFileInfo(filename string) (*mm_model.FileInfo, error) {
 	fileInfoID := getFileInfoID(parts[0])
 
 	if (fileInfoID) == "" {
-		return nil, nil
+		return nil, model.NewErrNotFound("No file found with name" + filename)
 	}
 
 	fileInfo, err := a.store.GetFileInfo(fileInfoID)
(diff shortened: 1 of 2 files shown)


harshilsharma63 (approved):

harshilsharma63:
@Rajat-Dabade can you add the steps to reproduce this bug in the ticket?

server/app/files.go
```diff
@@ -60,7 +60,16 @@ func (a *App) GetFileInfo(filename string) (*mm_model.FileInfo, error) {
...
-	fileInfoID := getFileInfoID(strings.Split(filename, ".")[0])
+	parts := strings.Split(filename, ".")
+	if len(parts) < 1 || len(parts[0]) <= 1 {
+		return nil, model.NewErrNotFound("No file found with name" + filename)
+	}
+
+	fileInfoID := getFileInfoID(parts[0])
+
+	if (fileInfoID) == "" {
+		return nil, nil
```
hmhealey, line 71: Should this case also return an error? I'd expect that this sort of method would always return either a FileInfo or an error, but this case returns nil for both of them

hmhealey (approved):
Looks good with one caveat

hmhealey:
Also, excellent explanation of your investigation here. I'm surprised that we keep the file extension on the uploaded file and that we have to calculate the file ID from the filename, so it's good to know that the code works that way

server/app/files.go
Rajat-Dabade: Done, returning an error with message `file not found`.

PR #35, opened 2024-11-11
MM-61537 fix for dark theme is boards

Rajat-Dabade:
#### Summary
Fix the dark theme for boards.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61537

Steps to reproduce this issue: 
https://www.loom.com/share/301b7127b6204553a0a81948585ed059

After fix: 

https://github.com/user-attachments/assets/fa7b37ce-b05c-4def-bec9-c069cf49e3a7


harshilsharma63 (approved):

yasserfaraazkhan (approved):
verified with steps Asaad shared. 
The UI in dark the works well
image

asaadmahmood (approved):

webapp/src/theme.ts
```diff
@@ -124,66 +122,51 @@ export function setTheme(theme: Theme | null): Theme {
...
+    const style = document.documentElement.style
+
+    style.setProperty('--center-channel-bg-rgb', style.getPropertyValue('--center-channel-bg-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--center-channel-color-rgb', style.getPropertyValue('--center-channel-color-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--button-bg-rgb', style.getPropertyValue('--button-bg-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--button-color-rgb', style.getPropertyValue('--button-color-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--sidebar-bg-rgb', style.getPropertyValue('--sidebar-bg-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--sidebar-text-rgb', style.getPropertyValue('--sidebar-text-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--link-color-rgb', style.getPropertyValue('--link-color-rgb') || consolidatedTheme.mainBg)
+    style.setProperty('--sidebar-text-active-border-rgb', style.getPropertyValue('--sidebar-text-active-border-rgb') || consolidatedTheme.mainBg)
```
hmhealey, line 149: As per the comment above, these variables are already set by the web app at this point, so this should only be needed for tests. Could we completely skip this when running outside of the tests?

Ideally, I think we'd do this in the Jest setup, but that seems like it might be a lot of work, so I think we could just wrap this block with an `if (process.env.NODE_ENV === 'test') {...}`

hmhealey (changes requested):

hmhealey (approved):

PR #40, opened 2024-11-25
MM-61945: Able to add individual users to the boards

Rajat-Dabade:
#### Summary
When adding individual members to the boards which were not part of the already linked boards was responded with 
`{error: "internal server error", "errorCode": 500}`. This was due to the server panicking if board members were not found. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61945


yasserfaraazkhan (approved):
Thank you Rajat, 
Verified below scenarios

- Admin creates and links channel to board
- User can access
- Admin adds users to channel and users can access,
- user can share the board/add other users
- Change the permission of the user and check if the flow is correct using the latest changed permission.
- Remove channel and add individual user and check the access.
- Remove the user and check the access to the board.

harshilsharma63 (approved):

PR #51, opened 2025-01-08
Upgraded workflows dependencies

saturninoabril:
#### Summary
As reported here - https://hub.mattermost.com/private-core/pl/a1xx1tmq6jng3qz9o9t1kdkzoy, where it failed in scorecard analysis.

Upgraded not only the action that caused failing job but all dependencies in the workflows.


mvitale1989 (approved):
Thanks for fixing Saturn!

PR #52, opened 2025-01-08
Migrate scorecard-analysis workflow to job token

mvitale1989:
#### Summary

Adapt the scorecard-analysis workflow to match the one used by the [scorecard repository itself](https://github.com/ossf/scorecard/blob/v5.0.0/.github/workflows/scorecard-analysis.yml).

[Here](https://github.com/ossf/scorecard-action?tab=readme-ov-file#breaking-changes-in-v2) are the scorecard-action docs, mentioning which permissions should the job token have, if `publish_results: true` is used


saturninoabril (approved):
Thanks @mvitale1989!

PR #58, opened 2025-01-23
 Apply and enforce file-level license headers

lieut-data:
#### Summary
Apply and enforce file-level license headers

Use `mattermost-govet` and `eslint`'s `header/header` to ensure each file begins with the appropriate header, then fix all instances for built source code. (e2e code excepted)

```
// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
// See LICENSE.txt for license information.
```

Note that this renames `LICENSE` to `LICENSE.txt` purely to align on the copyright messaging across repos. Note also that a number of preliminary golangci-lint fixes were required just to get this to build.

#### Ticket Link
None


alerque (changes requested):
Per-file license headers are great, but they should be made with SPDX format tags.

lieut-data:
Thanks, @alerque -- it's definitely something we'll consider for a future revision. For now, we're converging on the pattern we already use within our core product as we bring our pre-packaged plugins inline.

.config/notice-file/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+We are automatically generating Notice.txt by using first-level dependencies of the project. The related pipeline uses `config.yaml` stored in this folder.
+
+
+## Configuration
+
+Sample:
+
+```
+title: "Mattermost Playbooks"
```
Rajat-Dabade, line 11: Mattermost Playbooks? It should be Mattermost Boards Plugin right?

.config/notice-file/README.md
```diff
@@ -0,0 +1,28 @@
...
+copyright: "©2015-present Mattermost, Inc.  All Rights Reserved.  See LICENSE for license information."
+description: "This document includes a list of open source components used in Mattermost Playbooks, including those that have been modified."
```
Rajat-Dabade, line 13: Same here.

Makefile
```diff
@@ -82,8 +82,11 @@ ifneq ($(HAS_SERVER),)
 		echo "golangci-lint is not installed. Please see https://github.com/golangci/golangci-lint#install-golangci-lint for installation instructions."; \
 		exit 1; \
 	fi; \
+
 	@echo Running golangci-lint
-	cd server && golangci-lint run ./...
+	# cd server && golangci-lint run ./...
```
Rajat-Dabade, line 87: Remove comments.

.config/notice-file/README.md
lieut-data: This is the "sample" configuration from the notice generator. It's debatable if we need to repeatedly document how the tool works, but for now I've kept it consistent with other plugins.

Makefile
lieut-data: Oops, thanks for catching! I've been having trouble running golangci-lint in this repository and shouldn't have committed this.

.config/notice-file/README.md
lieut-data: See above, but basically this is a "sample configuration".

PR #64, opened 2025-02-04
Updated template loading validation

harshilsharma63:
#### Summary
The root cause was that the validations which were added for normal operation were incorrect for when the initial  templates were loaded on a fresh installation.

As a fix, I've separated the checks and created separate function for performing the special checks which are only valid during an import.

Cases tested-

1. Fresh boards install with no prior boards database tables
2. Upgrading an existing boards installation
3. Image upload
4. File attachments
5. Markdown in comments
6. Exporting and importing board archives
7. Managing categories and boards

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-62852


yasserfaraazkhan (approved):

amyblais:
@mgdelacroix Please help prioritize this PR review.

PR #65, opened 2025-02-04
Fixed path in command

harshilsharma63:
#### Summary
Used correct variable in make dist commands to construct dist dir path

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-62870


Rajat-Dabade (approved):

lieut-data (approved):
Aha, thanks for catching this! I missed this "non-idiomatic" pattern in boards relative to other plugins. Maybe an opportunity to align in the future?

PR #71, opened 2025-02-07
Removed Metropolis font import

marianunez:
#### Summary

Boards was importing a different version of Metropolis that was causing conflict with the webapp imported fonts. We no longer need to import it in Boards as it will always run as a plugin and can expect it to have it loaded in the Webapp. 

This import was from legacy Focalboard when running standalone. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62774


yasserfaraazkhan (approved):
- with plugins installed, the font appears correctly when we search with underscore.
- Able to create boards, add comments to card, move cards.

hmhealey (approved):

Rajat-Dabade (approved):
LGTM 🚀

PR #76, opened 2025-02-17
Bumped  mattermost/server/public/model version

marianunez:
#### Summary

Bumped  mattermost/server/public/model version to v0.1.10

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63191


.github/workflows/codeql-analysis.yml
```diff
@@ -26,20 +26,20 @@ jobs:
...
     
     # Autobuild attempts to build any compiled languages
     - name: Autobuild
-      uses: github/codeql-action/autobuild@v3.27.6
+      uses: github/codeql-action/autobuild@v3
 
     # Perform Analysis
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v3.27.6
+      uses: github/codeql-action/analyze@v3
```
marianunez, line 45: @saturninoabril @yasserfaraazkhan I had to do this change (thanks Copilot! 😄) because the pipeline was failing to finding the version `v3.27.6`. Is this version good or did we have a reason we were sticking to the specific version?

Rajat-Dabade (approved):
🚀 🚀 🚀

.github/workflows/codeql-analysis.yml
yasserfaraazkhan: V3 should work @marianunez. It will get the latest version.
image

yasserfaraazkhan (approved):

.github/workflows/codeql-analysis.yml
enzowritescode: For security reasons and to guarantee reproducible results, we always pin the version.
That way there are no unexplainable bugs if something is using a different version from one run to the next, or if a vulnerability or otherwise is introduced into a new version.

enzowritescode (changes requested):

.github/workflows/codeql-analysis.yml
marianunez: Makes sense, @enzowritescode @toninis any recomendations on what version we should pin this to?

enzowritescode: @marianunez it looks like github/codeql-action/init@v3.28.9 is the version being used in mattermost/mattermost which appears to be the latest 3.X version. There are some inconsistencies across the board though (see link below). Including @toninis for awareness.

https://github.com/search?q=org%3Amattermost%20github%2Fcodeql-action%2Finit&type=code

I suggest we set it to 3.28.9 and later we can make it consistent across the board in the other repositories.

enzowritescode (approved):

harshilsharma63 (approved):

PR #79, opened 2025-02-21
Remove unused dependencies

enzowritescode:
#### Summary

TLDR: removed unused dependencies

Long story:
I was looking into some dependencies that could be upgraded and the `package.json` dependencies seemed to be off and I couldn't find how some of the dependencies were used. I ran [depcheck](https://www.npmjs.com/package/depcheck) and it confirmed that some dependencies were unused.

#### Ticket Link

NA


marianunez (approved):

Rajat-Dabade (approved):

harshilsharma63 (approved):
I tried the app and it works. One more thing we can do is to remove the notice for deleted dependencies from `NOTICE.txt`.

PR #95, opened 2025-05-04
MM-63181 Fix number formatting issue in Boards Plugin for CSV export

Rajat-Dabade:
#### Summary
You can enter numbers with commas, for example, as thousand or decimal separators depending on your locale, into number-type columns. But when exporting to CSV, those values were being treated as numbers, which sometimes led to issues like `NaN` showing up or the values getting split across multiple columns. This change fixes that by treating the values as plain text during export, so they’re preserved exactly as entered.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63181

### DIFF
--- a/webapp/src/csvExporter.ts
+++ b/webapp/src/csvExporter.ts
@@ -87,7 +87,12 @@ class CsvExporter {
                 if (property.type === 'updatedBy') {
                     propertyValue = card.modifiedBy
                 }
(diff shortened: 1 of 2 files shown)


harshilsharma63 (approved):
Tested, works fine!

yasserfaraazkhan (approved):

PR #99, opened 2025-05-15
Added check for NaN in kanban column header

Rajat-Dabade:
#### Summary
I wasn’t able to reproduce the original issue https://mattermost.atlassian.net/browse/MM-63231, but I added a check to handle cases where the code might return a `NaN` value.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63231

### DIFF
--- a/webapp/src/components/calculations/calculations.ts
+++ b/webapp/src/components/calculations/calculations.ts
@@ -143,39 +143,41 @@ function sum(cards: readonly Card[], property: IPropertyTemplate): string {
 
     cardsWithValue(cards, property).
         forEach((card) => {
-            result += parseFloat(getCardProperty(card, property) as string)
+            const value = parseFloat(getCardProperty(card, property) as string)
+            if (!isNaN(value)) {
+                result += value
+            }
         })
 
     return String(Utils.roundTo(result, ROUNDED_DECIMAL_PLACES))
 }
 
 function average(cards: readonly Card[], property: IPropertyTemplate): string {
-    const numCards = cardsWithValue(cards, property).length
-    if (numCards === 0) {
+    const validCards = cardsWithValue(cards, property).filter(card => {
+        const value = parseFloat(getCardProperty(card, property) as string)
+        return !isNaN(value)
+    })
+    
+    if (validCards.length === 0) {
         return '0'
     }
 
-    const result = parseFloat(sum(cards, property))
-    const avg = result / numCards
+    const result = parseFloat(sum(validCards, property))
+    const avg = result / validCards.length
     return String(Utils.roundTo(avg, ROUNDED_DECIMAL_PLACES))
 }
 
 function median(cards: readonly Card[], property: IPropertyTemplate): string {
-    const sorted = cardsWithValue(cards, property).
-        sort((a, b) => {
-            if (!getCardProperty(a, property)) {
-                return 1
-            }
+    const validCards = cardsWithValue(cards, property).filter(card => {
+        const value = parseFloat(getCardProperty(card, property) as string)
+        return !isNaN(value)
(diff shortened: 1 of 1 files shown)


yasserfaraazkhan (approved):

harshilsharma63 (approved):

webapp/src/components/calculations/calculations.ts
```diff
@@ -143,39 +143,41 @@ function sum(cards: readonly Card[], property: IPropertyTemplate): string {
...
+            }
         })
 
     return String(Utils.roundTo(result, ROUNDED_DECIMAL_PLACES))
 }
 
 function average(cards: readonly Card[], property: IPropertyTemplate): string {
-    const numCards = cardsWithValue(cards, property).length
-    if (numCards === 0) {
+    const validCards = cardsWithValue(cards, property).filter(card => {
```
hmhealey, line 156: 0/5 on if this actually makes sense to change since I don't know where else `cardsWithValue` is used, but it seems like it could make sense to have `cardsWithValue` filter the result before its returned to avoid having to repeat this logic

webapp/src/components/calculations/calculations.ts
```diff
@@ -143,39 +143,41 @@ function sum(cards: readonly Card[], property: IPropertyTemplate): string {
...
+        return !isNaN(value)
+    })
+    
+    if (validCards.length === 0) {
         return '0'
     }
 
-    const result = parseFloat(sum(cards, property))
-    const avg = result / numCards
+    const result = parseFloat(sum(validCards, property))
```
hmhealey, line 165: Not something that needs to change here since it's existing behaviour, but the fact that these functions return strings instead of numbers seems very odd. Some combination of those might be how a `NaN` value slipped in to be the root cause of this bug.

I assume there's a good reason for it to return a string (probably for supporting data types of other fields), but the existing logic does look a bit weird to me

hmhealey (approved):
Looks good to me despite my confusion over how these calculation functions are currently passing around strings 😅

webapp/src/components/calculations/calculations.ts
Rajat-Dabade: Good point. Right now, `cardsWithValue` just checks for the existence of a value, not whether it’s a valid number. Since not all properties are numeric (e.g., text, date, etc.), I kept the `isNaN` filtering separate to avoid introducing type-specific logic into `cardsWithValue`, which is currently generic.

That said, if we find ourselves repeating numeric validation in multiple places, it might make sense to extract a `numericCardsWithValue` helper instead.

webapp/src/components/calculations/calculations.ts
Rajat-Dabade: Yeah, I had the same thought when I first looked at it. I think the reason `sum` returns a string is mainly for formatting for display purposes. Some of the other property types probably expect a string, so that might be why it’s done this way.

I agree it does feel a bit weird to convert it back to a number with `parseFloat(sum())` just to do math. It might be worth revisiting at some point to return a number by default and only format it at the UI level. But for now, I just kept the existing behaviour to avoid unexpected side effects.

Definitely open to cleaning it up later if it makes sense!

hmhealey: I think you're right about it being for display because it rounds off extra decimal places.

And yeah, I'm 0/5 on if we ever need to revisit this. The existing code is a bit odd, but changing it is pretty low impact unless we find it causes more issues

Rajat-Dabade: Yup, I think we need a revisit on this. Maybe rewrite some old code if need.

PR #102, opened 2025-05-21
[MM-63260] Fixed comments select text under each card - issue #63

### LINKED ISSUE #63: Bug: No ability to copy comments
Hi!
The ability to copy text from comments in cards is gone. Also, copying text via the right mouse button does not work. This option was available before.

iosifpeterfi:
#### Summary

This allows users to select text under comments on the board

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-boards/issues/63
  Fixes [MM-63260](https://mattermost.atlassian.net/browse/MM-63260)

[MM-63260]: https://mattermost.atlassian.net/browse/MM-63260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/src/components/cardDetail/comment.scss
+++ b/webapp/src/components/cardDetail/comment.scss
@@ -67,5 +67,6 @@
 
     .comment-markdown > * {
         white-space: pre-wrap;
+        user-select: auto;
     }
 }


harshilsharma63 (approved):

webapp/src/components/cardDetail/comment.scss
```diff
@@ -67,5 +67,6 @@
 
     .comment-markdown > * {
         white-space: pre-wrap;
+        user-select: auto;
```
Rajat-Dabade, line 70: Why not `user-select: text` if we are sure about allowing the user to select the text inside the element?
`user-select: auto` leaves the decision up to the browser’s default behaviour or what’s inherited from parent elements.

Rajat-Dabade (changes requested):

yasserfaraazkhan (approved):

webapp/src/components/cardDetail/comment.scss
iosifpeterfi: I agree, is probably best to have it as 'text', since 'auto' can have different outcome.

peetCreative:
This was solved here, can be closed. @iosifpeterfi 
https://github.com/mattermost/mattermost-plugin-boards/commit/6e90b8dc90a459b8fd50301fad9be3199dff3602

PR #103, opened 2025-05-21
Fix search in the Person field

Rajat-Dabade:
#### Summary
The search in the person field was case-sensitive, causing inconsistent results. For example, searching "harshil" worked, but "Harshil" did not. This PR resolves the issue by making the search case-insensitive.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64373


yasserfaraazkhan (approved):

hmhealey (approved):

Rajat-Dabade:
@harshilsharma63 You're right, thinking about it now, the changes should be made on the server side rather than the client side. I will update the code. Nice catch 🙌 .

Rajat-Dabade:
@yasserfaraazkhan Can you please re-test this, made changes in the server side, and reverted from the client side.

PR #108, opened 2025-06-16
Improve file path validation and error handling

Rajat-Dabade:
#### Summary
- Added input validation to file path components
- Fix silent failing in `getDestinationFilePath` function to return proper errors.
- Additionally, enhance error handling across file operations in other functions in `files.go`.
- Added validations to ensure path components contain only valid characters. 
- Updated tests    

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64596

### DIFF
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/mattermost/mattermost-plugin-boards/server/model"
@@ -35,7 +36,10 @@ func (a *App) SaveFile(reader io.Reader, teamID, boardID, filename string, asTem
 	if asTemplate {
 		newFileName = filename
 	}
-	filePath := getDestinationFilePath(asTemplate, teamID, boardID, newFileName)
+	filePath, pathErr := getDestinationFilePath(asTemplate, teamID, boardID, newFileName)
+	if pathErr != nil {
+		return "", fmt.Errorf("invalid file path parameters: %w", pathErr)
+	}
 
 	fileSize, appErr := a.filesBackend.WriteFile(reader, filePath)
 	if appErr != nil {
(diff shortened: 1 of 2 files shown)


server/app/files.go
```diff
@@ -22,6 +23,10 @@ const emptyString = "empty"
 
 var errEmptyFilename = errors.New("IsFileArchived: empty filename not allowed")
 var ErrFileNotFound = errors.New("file not found")
+var errEmptyPathComponent = errors.New("empty path component")
```
lieut-data: Given it doesn't look like we use (or intend to use) these anywhere outside of returning them, should we just inline? (If it was exported, that would be a different matter, but then I'd probably argue for a simpler base error instead that's wrapped.)

server/app/files.go
```diff
@@ -119,20 +127,66 @@ func (a *App) GetFilePath(teamID, rootID, fileName string) (*mm_model.FileInfo,
...
+		return errEmptyPathComponent
+	}
+
+	if strings.Contains(component, "..") {
+		return errInvalidPathComponent
+	}
+
+	if strings.HasPrefix(component, "/") || strings.HasPrefix(component, "\\") {
+		return errInvalidPathFormat
+	}
```
lieut-data: Is this redundant, given the regex below?

server/app/files.go
```diff
@@ -119,20 +127,66 @@ func (a *App) GetFilePath(teamID, rootID, fileName string) (*mm_model.FileInfo,
...
 	}
-	return filepath.Join(utils.GetBaseFilePath(), filename)
+	return filepath.Join(utils.GetBaseFilePath(), filename), nil
+}
+
+// validatePathComponent ensures a path component contains only valid characters.
+func validatePathComponent(component string) error {
+	if component == "" {
+		return errEmptyPathComponent
+	}
```
lieut-data: Is this redundant, given the regex below?

server/app/files.go
```diff
@@ -119,20 +127,66 @@ func (a *App) GetFilePath(teamID, rootID, fileName string) (*mm_model.FileInfo,
...
+
+	if strings.Contains(component, "..") {
+		return errInvalidPathComponent
+	}
```
lieut-data: Given we don't allow path separators, do we need this check at all? (Is there anything wrong with a filename with multiple periods?)

lieut-data (approved):

server/app/files.go
```diff
@@ -119,20 +127,66 @@ func (a *App) GetFilePath(teamID, rootID, fileName string) (*mm_model.FileInfo,
 	if fileInfo != nil && fileInfo.Path != "" && fileInfo.Path != emptyString {
 		filePath = fileInfo.Path
 	} else {
+		// Validate path components to ensure proper file path handling
+		if err := validatePathComponent(teamID); err != nil {
```
enzowritescode: I believe this validation could be more strict, ex: https://github.com/mattermost/mattermost/blob/548a47ae566912d1ac6f169145bfb0c534986d11/server/public/model/utils.go#L768

server/app/files.go
enzowritescode: Technically a filename can have multiple dots under certain circumestances such as `myfile....example.json`, although there isn't really a need to allow it.

server/app/files.go
enzowritescode: Agreed, that way we don't have unnecessary variables in memory

enzowritescode (changes requested):
Every place we are checking for an ID (teamID, boardID, channelID, etc) we can perform more strict validation. Example: https://github.com/mattermost/mattermost/blob/548a47ae566912d1ac6f169145bfb0c534986d11/server/public/model/utils.go#L768

server/app/files.go
harshilsharma63: We can call `mmModal.IsValidId()` for team ID and channel ID as they are Mattermost's team and channel ID, for everything else, `validatePathComponent` is fine.

enzowritescode: Agreed

server/app/files.go
```diff
@@ -119,20 +127,66 @@ func (a *App) GetFilePath(teamID, rootID, fileName string) (*mm_model.FileInfo,
...
+	}
+
 	// if saving a file for a template, save using the "old method" that is /teamID/boardID/fileName
 	// this will prevent template files from being deleted by DataRetention,
 	// which deletes all files inside the "date" subdirectory
 	if isTemplate {
-		return filepath.Join(teamID, boardID, filename)
+		// Use a safe base path for templates with proper validation
+		basePath := filepath.Join(utils.GetBaseFilePath(), "templates")
+		return filepath.Join(basePath, teamID, boardID, filename), nil
```
esarafianou: This seems to be unintended based on the comment above. The reason we hadn't used  the `utils.GetBaseFilePath()` before was exactly what the comment says: Prevent template files from being deleted by DataRetention. They will now be deleted.

esarafianou (changes requested):

server/app/files.go
daw1012345: Just wanted to chime in that `strings.Contains(component, "..")` will reject `myfile....example.json` since it does contain the dots: `myfile**..**..example.json`. If the goal is to allow filenames like this, then this won't work.

server/app/files.go
Rajat-Dabade: This change was because of the linter error `do not define dynamic errors, use wrapped static errors instead: "fmt.Errorf(\"invalid teamID in GetFilePath: %s\", teamID)`, but I agree with the unnecessary variable in memory. It's better to disable the linter check for these lines.

server/app/files.go
Rajat-Dabade: Removed it. Nice catch 🙌

server/app/files.go
Rajat-Dabade: Removed it.

server/app/files.go
Rajat-Dabade: Agreed, made the changes.

server/app/files.go
Rajat-Dabade: Yes, this is a nice catch @esarafianou, I have reverted the changes for the base path.

esarafianou (approved):

server/app/files.go
```diff
@@ -119,20 +123,53 @@ func (a *App) GetFilePath(teamID, rootID, fileName string) (*mm_model.FileInfo,
...
 	return fileInfo, filePath, nil
 }
 
-func getDestinationFilePath(isTemplate bool, teamID, boardID, filename string) string {
+func getDestinationFilePath(isTemplate bool, teamID, boardID, filename string) (string, error) {
+	// Validate inputs to ensure proper file path handling
+	if !mm_model.IsValidId(teamID) {
+		return "", fmt.Errorf("getDestinationFilePath: invalid team ID for teamID %s", teamID) //nolint:err113
+	}
+	if err := validatePathComponent(boardID); err != nil {
```
enzowritescode: This can use `IsValidId()`

enzowritescode (changes requested):

server/app/files.go
Rajat-Dabade: True, Updated the code.

PR #113, opened 2025-07-07
 Fix for Global Template Initialisation

Rajat-Dabade:
#### Summary
Global template initialisation failed after #108 due to overly strict validation in `getDestinationFilePath()` that rejected team ID "0" (`GlobalTeamID`).
`Error: "cannot initialise global templates for team 0: invalid team ID for team 0"`
Impact: Only 3 templates were imported instead of 13 on fresh installations

#### Solution
Modified `getDestinationFilePath()` to allow `GlobalTeamID` ("0") for template operations while maintaining validation for other use cases:
```go
if !mm_model.IsValidId(teamID) && !(isTemplate && teamID == model.GlobalTeamID) {
    return "", fmt.Errorf("getDestinationFilePath: invalid team ID for teamID %s", teamID)
}
```

https://hub.mattermost.com/private-core/pl/fy98hhctff8fbf7ecnpx1qqcsa

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64688

### DIFF
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -180,20 +180,18 @@ func getFileInfoID(fileName string) string {
 
 func (a *App) GetFileReader(teamID, boardID, filename string) (filestore.ReadCloseSeeker, error) {
 	// Validate path components - be restrictive here since this can be called from API handlers
-	// For GlobalTeamID, verify that it's actually for a template board
-	if !mm_model.IsValidId(teamID) {
-		if teamID == model.GlobalTeamID {
-			// Only allow GlobalTeamID if this is actually a template board
-			board, err := a.GetBoard(boardID)
-			if err != nil {
-				return nil, fmt.Errorf("GetFileReader: cannot validate board for GlobalTeamID: %w", err)
-			}
-			if !board.IsTemplate {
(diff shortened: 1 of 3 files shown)


server/app/files.go
```diff
@@ -141,7 +141,8 @@ func (a *App) GetFilePath(teamID, boardID, fileName string) (*mm_model.FileInfo,
 
 func getDestinationFilePath(isTemplate bool, teamID, boardID, filename string) (string, error) {
 	// Validate inputs to ensure proper file path handling
-	if !mm_model.IsValidId(teamID) {
+	// Only allow GlobalTeamID for template operations to prevent path traversal attacks
+	if !mm_model.IsValidId(teamID) && !(isTemplate && teamID == model.GlobalTeamID) {
```
harshilsharma63: Whenever we will validate team ID, we will have to handle the global team ID case for templates. So, it would be better if we create a function that encapsulates the whole logic of validating team IDs, something like `ValidateTeamId(teamId, isTemplate)`.

harshilsharma63 (changes requested):

server/app/files.go
Rajat-Dabade: Done.

hmhealey (approved):

PR #114, opened 2025-07-08
Added file ownership validation to file access endpoints

Rajat-Dabade:
### Summary
Adds validation to ensure files can only be accessed through their correct board and team context in file serving endpoints.

#### Problem: 
The file access endpoints (`/files/teams/{teamID}/{boardID}/{filename}` and `/files/teams/{teamID}/{boardID}/{filename}/info`) were not verifying that the requested file belonged to the specified board. This could lead to inconsistent behaviour when accessing files across different board contexts.

#### Solution: 
Added file ownership validation: New `ValidateFileOwnership()` function ensures files belong to the requested board and team
Dual validation approach:
For files with explicit `team/board` paths, validates the path structure
For files with newer storage format: validates that the file is referenced by blocks within the specified board
API-level protection: Added validation to both file-serving and file-info endpoints.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64660

### DIFF
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -110,6 +110,8 @@ func (a *App) ValidateFileOwnership(teamID, boardID, filename string) error {
 }
 
 // validateFileReferencedByBoard checks if a file is referenced by blocks in the specified board.
+// Files use different storage patterns (teamID/boardID/filename for templates, boards/YYYYMMDD/filename for regular files).
+// Path mismatches don't indicate malicious files, just different storage patterns.
 func (a *App) validateFileReferencedByBoard(boardID, filename string) error {
 	imageBlocks, err := a.store.GetBlocksWithType(boardID, model.TypeImage)
 	if err != nil {
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -111,19 +111,33 @@ func (a *App) ValidateFileOwnership(teamID, boardID, filename string) error {
 
 // validateFileReferencedByBoard checks if a file is referenced by blocks in the specified board.
 func (a *App) validateFileReferencedByBoard(boardID, filename string) error {
-	blocks, err := a.GetBlocksForBoard(boardID)
+	imageBlocks, err := a.store.GetBlocksWithType(boardID, model.TypeImage)
 	if err != nil {
 		return err
 	}
 
-	for _, block := range blocks {
-		if block.Type == model.TypeImage || block.Type == model.TypeAttachment {
-			if fileID, ok := block.Fields[model.BlockFieldFileId].(string); ok && fileID == filename {
-				return nil
-			}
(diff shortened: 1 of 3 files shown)


yasserfaraazkhan (approved):

enzowritescode (approved):

server/app/files.go
```diff
@@ -86,7 +87,59 @@ func (a *App) GetFileInfo(filename string) (*mm_model.FileInfo, error) {
...
+	fileInfo, err := a.GetFileInfo(filename)
+	if err != nil {
+		return err
+	}
+	if fileInfo != nil && fileInfo.Path != "" && fileInfo.Path != emptyString {
+		expectedPath := filepath.Join(teamID, boardID, filename)
+		if fileInfo.Path == expectedPath {
+			return nil
+		}
+		if err := a.validateFileReferencedByBoard(boardID, filename); err != nil {
```
harshilsharma63, line 101: Can you explain why this check is needed if the above expected file patch check is `false`? `validateFileReferencedByBoard` fetches all blocks of a board and iterates through them which would involve fetching a large amount of data from database and processing it. Any way we can avoid it?

Rajat-Dabade: This is needed because of the evolution of file storage patterns. Files are being stored in different file patterns like: 
`teamID/boardID/filename` for template files and `boards/YYYYMMDD/filename` for regular files. Due to path validation, if the `fileInfo.Path` doesn't match the expected `teamID/boardID/filename` format, it doesn't necessarily mean the file is malicious, it could be a legitimate file using a different storage pattern.


You're right about the performance concern. I have made some changes that can reduce some database load. Rather than querying the database with `GetBlockForBoards`, querying the database with `GetBlocksWithType ' is more optimal. 

Before 

```go
blocks, err := a.GetBlocksForBoard(boardID)  // Fetches ALL blocks for the board
```

AFter 
```go
imageBlocks, err := a.store.GetBlocksWithType(boardID, model.TypeImage)
attachmentBlocks, err := a.store.GetBlocksWithType(boardID, model.TypeAttachment)
```

harshilsharma63: This is fine for now but this does add a significant performance penalty in media-heavy boards.

We need to scan this way because we don't store the board ID in `fileinfo`, so there is no way to know which board a file belong to without scanning the board's blocks.

Can you file a ticket for future to add a data migration to store board ID in `fileInfo` table to optimize it?

And it will be very helpful if you can add the explanation you added in your comment as a code comment as well.

harshilsharma63 (approved):

server/app/files.go
Rajat-Dabade: Added the comment and created the ticket: https://mattermost.atlassian.net/browse/MM-64852

PR #115, opened 2025-07-09
Re-import template on new instance and template file access issue

Rajat-Dabade:
### Summary
This PR fixes the following:
- All default templates will be restored after server restart
- Templates with files (Content Calendar, Personal Tasks, Roadmap, Sprint Planner, Welcome to Boards) will work correctly when creating new boards
- The system properly validates template file operations while maintaining security

### Issue
1. Template Re-import Issue
Problem: Only 3 templates were showing because the template initialisation system wasn't re-running after the regression was fixed.
Solution: Bumped `defaultTemplateVersion` from 6 to 7 in `server/app/templates.go`. This forces the system to detect that existing templates have an "old" version and triggers a complete re-import of all default templates.

2. Template File Access Issue
Problem: Templates with files/images were failing to create boards due to GetFilePath not allowing teamID = "0" (GlobalTeamID).
Solution: Updated the GetFilePath function in server/app/files.go to properly handle GlobalTeamID for templates.

### Ticket Link
https://mattermost.atlassian.net/browse/MM-64800

### DIFF
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -124,8 +124,18 @@ func (a *App) GetFilePath(teamID, boardID, fileName string) (*mm_model.FileInfo,
 		filePath = fileInfo.Path
 	} else {
 		// Validate path components to ensure proper file path handling
-		if !mm_model.IsValidId(teamID) {
-			return nil, "", fmt.Errorf("GetFilePath: invalid team ID for teamID %s", teamID) //nolint:err113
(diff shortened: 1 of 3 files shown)


Rajat-Dabade:
@yasserfaraazkhan This is not a blocker. It's just `React` complaining about following some rules. It's okay to have this error for now.

yasserfaraazkhan (approved):

devinbinnie (approved):

PR #116, opened 2025-07-09
Fix the unlinking issue of board to channel

Rajat-Dabade:
#### Summary
This PR fix the board unlinking issue with the channel. 

Problem: 
The issue was in the `BoardPatch.IsValid()` method in `server/model/board.go`. When unlinking a board from a channel, the API sends a patch with `ChannelID` set to an empty string "", but the validation logic wasn't handling this correctly. Updated the method to handle an empty `ChannelID`. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64805


pvev (approved):
🚀

PR #126, opened 2025-08-19
Send csrf token

harshilsharma63:
#### Summary
Used the MM redux `Client4.getOptions()` method to get CSRF token added to requests.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-65041

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -140,9 +140,6 @@ func (a *API) requireCSRFToken(next http.Handler) http.Handler {
 
 func (a *API) checkCSRFToken(r *http.Request) bool {
 	token := r.Header.Get(HeaderRequestedWith)
-
-	fmt.Println(fmt.Sprintf("############################## token: %s", token))
-
 	return token == HeaderRequestedWithXML
 }
 


harshilsharma63:
> I think the CSRF token also needs to be added to `uploadAttachment`

Great catch @enzowritescode ! I've added it now.

webapp/src/octoClient.ts
```diff
@@ -69,11 +71,11 @@ class OctoClient {
     async login(username: string, password: string): Promise<boolean> {
         const path = '/api/v2/login'
         const body = JSON.stringify({username, password, type: 'normal'})
-        const response = await fetch(this.getBaseURL() + path, {
+        const response = await fetch(this.getBaseURL() + path, Client4.getOptions({
```
marianunez, line 74: How can we guarantee that future calls here properly use `Client4.getOptions`? We should consider creating a base fetch call that makese sure to pass through the `options` through the client4 as Playbooks is doing: https://github.com/mattermost/mattermost-plugin-playbooks/blob/4d758cb0e2288cd9e2097c43c32f1c20a8bb5108/webapp/src/client.ts#L724-L725

marianunez (changes requested):
LGTM -

webapp/src/octoClient.ts
enzowritescode: It would be nice if that sort of thing was in a library we could import to minimize the amount of copy-pasted code in plugins.

harshilsharma63: @marianunez I'd recommend postponing the base function implementation for now.

The current functions have inconsistent patterns - some return false/null/undefined on failures, others return empty arrays or promises. Standardizing this would require significant refactoring time with limited benefit, given this file is rarely modified (last changed 6 months ago, previous change was 2.5 years ago).

Without proper test coverage, I'm not comfortable making changes of this scope without an urgent need. Future developers will follow the existing patterns when making updates and since the new API will immediately fail even locally, the missing call to `Client4.getOptions` won't go unnoticed.

marianunez: Agreed that the lack of test coverage introduces a risk right now. Can we create a follow up ticket for this ?

marianunez (approved):
Thanks @harshilsharma63!

webapp/src/octoClient.ts
harshilsharma63: I've created a ticket here - https://mattermost.atlassian.net/browse/MM-65168

PR #131, opened 2025-09-23
Added checks for user and block for file and subscription

Rajat-Dabade:
#### Summary
This PR ensure that only authorised users can subscribe to a block (must have view access to the block of the board)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64971

### DIFF
--- a/server/api/subscriptions.go
+++ b/server/api/subscriptions.go
@@ -73,11 +73,6 @@ func (a *API) handleCreateSubscription(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	session := ctx.Value(sessionContextKey).(*model.Session)
 
-	auditRec := a.makeAuditRecord(r, "createSubscription", audit.Fail)
-	defer a.audit.LogRecord(audit.LevelModify, auditRec)
-	auditRec.AddMeta("subscriber_id", sub.SubscriberID)
-	auditRec.AddMeta("block_id", sub.BlockID)
-
 	// User can only create subscriptions for themselves (for now)
 	if session.UserID != sub.SubscriberID {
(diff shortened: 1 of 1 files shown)


yasserfaraazkhan (approved):

server/api/subscriptions.go
```diff
@@ -84,14 +84,19 @@ func (a *API) handleCreateSubscription(w http.ResponseWriter, r *http.Request) {
...
-	_, bErr := a.app.GetBlockByID(sub.BlockID)
+	// check for valid block and permissions to view its board
+	block, bErr := a.app.GetBlockByID(sub.BlockID)
 	if bErr != nil {
 		message := fmt.Sprintf("invalid blockID: %s", bErr)
 		a.errorResponse(w, r, model.NewErrBadRequest(message))
 		return
 	}
 
+	if !a.permissions.HasPermissionToBoard(sub.SubscriberID, block.BoardID, model.PermissionViewBoard) {
```
harshilsharma63, line 90: Can we move this further up in this function to avoid creating audit records if the user doesn't have permission? We don't create audit records in other places for simple cases like missing permissions.

harshilsharma63, line 90: Is the same check also needed in `handleDeleteSubscription` and `handleGetSubscriptions` APIs?

Rajat-Dabade: Okay.

Rajat-Dabade: No, delete/get are self-only. User can not be able to delete or get other users' subscriptions, so no boards check is required here.

harshilsharma63 (approved):

PR #122, opened 2025-07-31
SQL query improvement

esarafianou:
#### Summary
Improve SQL query in `undeleteBlockChildren`

#### Ticket Link
N/A

PR #132, opened 2025-09-23
Fix: Error message modal pops up when hiding a Board

Rajat-Dabade:
##### Issue: 
- When a board is deleted, it is not removed from the sidebar category state. This results in, if someone tries to hide the board, the system still searches for that deleted board and attempts to load it, which opens a “Join Private Board” modal.
- Also, when loading a board, the app checks `localStorage` for `lastBoardId`. Since there was no validation to confirm whether the board still exists, sometimes it tried to open a deleted board.
- Another issue was that even if the REST API calls (for reordering boards or moving them between categories) failed, the client still updated the state, even if the API calls failed. This led to inconsistencies, after a hard refresh, the board position would change back to the original, ignoring the attempted change. The rollback mechanism was missing here.

##### Solution: 
This PR handles 
- Added Rollback for
  - Same category reordering: on REST failure, revert order.
  - Cross-category move: on move failure, revert both mapping and order; on reorder failure (post-move), revert order only.
- After Hide, choose the next board only from locally known, non-deleted boards
- If no valid boards, clear `lastBoardID` and route to template picker.
- If localStorage contains a stale board ID, then pick the first visible board from sidebar categories that exists in the local boards set and redirect there, updating lastBoardID. If not found, clear `lastBoardID` and redirect it to the template picker page. 

##### Impact
- Prevents stale board IDs from being present in sidebar state after boards are deleted.
- Prevents navigation to a board that has been deleted, whether from hide actions or from outdated localStorage references.
- Rollback if the API fails for reordering and category change. 

⸻

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65648

### DIFF
--- a/server/services/store/sqlstore/blocks.go
+++ b/server/services/store/sqlstore/blocks.go
@@ -1038,6 +1038,8 @@ func (s *SQLStore) undeleteBlockChildren(db sq.BaseRunner, boardID string, paren
 		return fmt.Errorf("undeleteBlockChildren unable to generate subquery: %w", err)
 	}
 
+	joinArgs := append([]interface{}{modifiedBy}, subQueryArgs...)
+
 	selectQuery := s.getQueryBuilder(db).
 		Select(
 			"bh.board_id",
@@ -1055,7 +1057,7 @@ func (s *SQLStore) undeleteBlockChildren(db sq.BaseRunner, boardID string, paren
 			"bh.created_by",
 		).
 		From(s.tablePrefix+"blocks_history AS bh").
-		InnerJoin("("+subQuerySQL+") AS sub ON bh.id=sub.id AND bh.insert_at=sub.max_insert_at", append(subQueryArgs, modifiedBy)...).
+		InnerJoin("("+subQuerySQL+") AS sub ON bh.id=sub.id AND bh.insert_at=sub.max_insert_at", joinArgs...).
 		Where(sq.NotEq{"bh.delete_at": 0})
 
 	columns := []string{


webapp/src/components/sidebar/sidebar.tsx
```diff
@@ -214,32 +212,34 @@ const Sidebar = (props: Props) => {
...
 
             dispatch(updateCategoryBoardsOrder({categoryID: toCategoryID, boardsMetadata: categoryBoardMetadata}))
 
-            const reorderedBoardIDs = categoryBoardMetadata.map((m) => m.boardID)
-            await octoClient.reorderSidebarCategoryBoards(team.id, toCategoryID, reorderedBoardIDs)
+            try {
+                const reorderedBoardIDs = categoryBoardMetadata.map((m) => m.boardID)
+                const updatedOrder = await octoClient.reorderSidebarCategoryBoards(team.id, toCategoryID, reorderedBoardIDs)
+                if (reorderedBoardIDs.length > 0 && updatedOrder.length === 0) {
+                    throw new Error('reorderSidebarCategoryBoards failed')
```
devinbinnie: Is this the only error we're catching? If so I think we can avoid the try/catch.
If not - what else are we expecting to go wrong here?

I'm wondering if this constitutes an error in the console or if it's just something we can fail silently on.

webapp/src/components/sidebar/sidebar.tsx
```diff
@@ -249,14 +249,31 @@ const Sidebar = (props: Props) => {
...
 
-            // optimistically updating the store to create a lag-free UI.
             await dispatch(updateCategoryBoardsOrder({categoryID: toCategoryID, boardsMetadata: categoryBoardMetadata}))
             dispatch(updateBoardCategories([{...fromCategoryBoardMetadata, categoryID: toCategoryID}]))
 
-            await mutator.moveBoardToCategory(team.id, boardID, toCategoryID, fromCategoryID)
+            try {
+                const moveResp = await octoClient.moveBoardToCategory(team.id, boardID, toCategoryID, fromCategoryID)
+                if (!moveResp || !moveResp.ok) {
+                    throw new Error('moveBoardToCategory failed')
```
devinbinnie: Wondering the same here.

webapp/src/store/sidebar.ts
```diff
@@ -175,6 +175,28 @@ const sidebarSlice = createSlice({
...
+            state.categoryAttributes = state.categoryAttributes.map((categoryBoards: CategoryBoards) => {
+                const filtered = categoryBoards.boardMetadata.filter((m) => !toRemove.has(m.boardID))
+                return {
+                    ...categoryBoards,
+                    boardMetadata: filtered,
+                }
+            })
+
+            // Recompute hiddenBoardIDs after pruning
+            state.hiddenBoardIDs = state.categoryAttributes.flatMap((ca) => ca.boardMetadata.reduce((collector, m) => {
```
devinbinnie: Can we not just use a `.filter` here?

webapp/src/components/sidebar/sidebar.tsx
Rajat-Dabade: Yup, that's the only error we're catching, so we can remove the try/catch.

webapp/src/components/sidebar/sidebar.tsx
Rajat-Dabade: Done, removed the try/catch.

webapp/src/store/sidebar.ts
Rajat-Dabade: Yes, we can use the `.filter` and `.map` combination to extract the boardID. I will make the changes.

webapp/src/components/sidebar/sidebar.tsx
```diff
@@ -249,14 +244,30 @@ const Sidebar = (props: Props) => {
...
+            if (!moveResp || !moveResp.ok) {
+                dispatch(updateCategoryBoardsOrder({categoryID: toCategoryID, boardsMetadata: previousToBoardsMetadata}))
+                dispatch(updateBoardCategories([{...fromCategoryBoardMetadata, categoryID: fromCategoryID}]))
+                return
+            }
+
+            try {
+                const reorderedBoardIDs = categoryBoardMetadata.map((m) => m.boardID)
+                const updatedOrder = await octoClient.reorderSidebarCategoryBoards(team.id, toCategoryID, reorderedBoardIDs)
+                if (reorderedBoardIDs.length > 0 && updatedOrder.length === 0) {
```
devinbinnie: Same try/catch change here?

devinbinnie (approved):
Approving to unblock, we should fix that other try/catch block as well to make it clearer.

webapp/src/components/sidebar/sidebar.tsx
Rajat-Dabade: Done.

webapp/src/components/sidebar/sidebar.tsx
```diff
@@ -214,32 +212,29 @@ const Sidebar = (props: Props) => {
...
             const categoryBoardMetadata = [...toSidebarCategory.boardMetadata]
             categoryBoardMetadata.splice(source.index, 1)
             categoryBoardMetadata.splice(destination.index, 0, toSidebarCategory.boardMetadata[source.index])
 
             dispatch(updateCategoryBoardsOrder({categoryID: toCategoryID, boardsMetadata: categoryBoardMetadata}))
 
             const reorderedBoardIDs = categoryBoardMetadata.map((m) => m.boardID)
-            await octoClient.reorderSidebarCategoryBoards(team.id, toCategoryID, reorderedBoardIDs)
+            const updatedOrder = await octoClient.reorderSidebarCategoryBoards(team.id, toCategoryID, reorderedBoardIDs)
+            if (reorderedBoardIDs.length > 0 && updatedOrder.length === 0) {
```
harshilsharma63, line 232: Can you add code comment what this condition means and why we need to handle it?

webapp/src/components/sidebar/sidebar.tsx
```diff
@@ -249,14 +244,26 @@ const Sidebar = (props: Props) => {
...
 
-            // optimistically updating the store to create a lag-free UI.
             await dispatch(updateCategoryBoardsOrder({categoryID: toCategoryID, boardsMetadata: categoryBoardMetadata}))
             dispatch(updateBoardCategories([{...fromCategoryBoardMetadata, categoryID: toCategoryID}]))
 
-            await mutator.moveBoardToCategory(team.id, boardID, toCategoryID, fromCategoryID)
+            // Persist the move; if request fails or server rejects, rollback silently
+            const moveResp = await octoClient
+                .moveBoardToCategory(team.id, boardID, toCategoryID, fromCategoryID)
+                .catch(() => undefined)
```
harshilsharma63: Log the error in catch.

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -139,6 +139,12 @@ const BoardPage = (props: Props): JSX.Element => {
                     boardId: activeBoardId,
                 }))
             }
+
+            // remove boards from all categories if they are deleted
+            const deletedBoardIds = teamBoards.filter((b: Board) => b.deleteAt && b.deleteAt !== 0).map((b) => b.id)
```
harshilsharma63, line 144: Is undo event handled for this? If I delete a board and undo, is the board restored into the right category?

webapp/src/components/sidebar/sidebar.tsx
Rajat-Dabade: Sure.

webapp/src/components/sidebar/sidebar.tsx
Rajat-Dabade: Sure.

webapp/src/pages/boardPage/boardPage.tsx
Rajat-Dabade: Will handle this in subsequent PR.

Rajat-Dabade:
@harshilsharma63 the undeleted of the board becomes inaccessible in an already existing issue. Created a separate ticket for it: https://mattermost.atlassian.net/browse/MM-66101

webapp/src/pages/boardPage/boardPage.tsx
harshilsharma63: A board becomes inaccessible can be handled in another PR, thats fine, but moving it to the right category should be handled in this PR itself as it is a bug that is introduced in this PR. On latest master, a deleted board, after un-deleting, shows up in the same category where it was before.

Rajat-Dabade: Yes. Done.

server/services/store/sqlstore/blocks.go
```diff
@@ -1055,7 +1057,7 @@ func (s *SQLStore) undeleteBlockChildren(db sq.BaseRunner, boardID string, paren
 			"bh.created_by",
 		).
 		From(s.tablePrefix+"blocks_history AS bh").
-		InnerJoin("("+subQuerySQL+") AS sub ON bh.id=sub.id AND bh.insert_at=sub.max_insert_at", append(subQueryArgs, modifiedBy)...).
+		InnerJoin("("+subQuerySQL+") AS sub ON bh.id=sub.id AND bh.insert_at=sub.max_insert_at", joinArgs...).
```
Rajat-Dabade, line 1060: The `modifiedBy` should be the first argument rather than 2nd in the `append` function because it should match the placeholder `?` order. This is the regression introduced in https://github.com/mattermost/mattermost-plugin-boards/pull/122/files. 

cc// @esarafianou

harshilsharma63 (approved):

server/services/store/sqlstore/blocks.go
esarafianou: Thanks Rajat!

PR #136, opened 2025-10-06
Fix: Attachment within cards are shown blank and not downloading

Rajat-Dabade:
#### Summary
Here is what is happening when deleting and undoing the attachment:
When the attachment was deleted: 
- The block was deleted and moved to the blocks_history table.
- The file was soft deleted (FileInfo.DeleteAt as set to the current timestamp).

When the deletion was undone: 
- The block was restored from the blocks_history table.
- The file was not restored - it remains soft deleted.
This is why only the block has been shown and not the file attachments.

Oct 4 Screenshot from Boards

This PR fixes this issue. If the block has `fileId` or `attachmentId`, then restore the file (i.e., set `FileInfo.DeleteAt` to 0). Also, there is a check if this same issue occurs when deleting the board and undoing it. For those who are already seeing this issue, if the block is still there and the associated file is already there in the database, then it should be restored. If the file is manually removed from the database, then do not return the block (do not show the block in the card).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65033


server/services/store/sqlstore/blocks.go
```diff
@@ -1038,6 +1062,8 @@ func (s *SQLStore) undeleteBlockChildren(db sq.BaseRunner, boardID string, paren
 		return fmt.Errorf("undeleteBlockChildren unable to generate subquery: %w", err)
 	}
 
+	joinArgs := append([]interface{}{modifiedBy}, subQueryArgs...)
```
mgdelacroix: nit: 

```suggestion
	joinArgs := append([]any{modifiedBy}, subQueryArgs...)
```

PR #140, opened 2025-10-09
Fix: Board redirect on itself issue

Rajat-Dabade:
#### Summary
I was not able to reproduce this issue locally, but suspect the following:
When we open the board in a new tab, and due to some reason, any error (this is still unknown, but might be related to the board not being found in the current team) occurs, this is what happens:
- Error occurs -> ErrorBoundary triggers
- Original code: Utils.getBaseURL(true) -> `http://domain/plugins/focalboard/error?id=unknown`
- Router sees /plugins/focalboard/error -> treats as legacy route
- Router redirects to /boards/error?id=unknown
- ErrorBoundary triggers again -> creates /plugins/focalboard/error?id=unknown
- Loop continues: /plugins/focalboard/error ->  /boards/error

Solution: 
Changing to `Utils.getFrontendBaseURL(true)` ensures:
- Error URL is /boards/error?id=unknown
- Router treats it as a modern route
- No legacy redirect occurs
- Loop stops

#### Ticket Link

## Summary

This PR fixes a critical redirect loop issue where opening a board in a new tab produces infinite redirects between `/plugins/focalboard/error` and `/boards/error`. The fix updates the ErrorBoundary component to use `getFrontendBaseURL(true)` instead of `getBaseURL(true)`, ensuring the error URL is treated as a modern route rather than triggering a legacy redirect.

**Key Changes:**

1. **Server Configuration** – Updated serverRoot construction from `baseURL + "/plugins/focalboard"` to `baseURL + "/boards"`

2. **Error Handling & Routing**
   - Error boundary now constructs redirect URLs using the correct frontend base URL
   - Added new `ViewNotFound` error case for invalid view IDs
   - Implemented runtime validations in `boardPage.tsx` to verify team, board, and view existence before loading
   - Dispatch appropriate errors ('team-undefined', 'board-not-found', 'view-not-found') when data is invalid

3. **Board & View Selection** – Refactored `teamToBoardAndViewRedirect.tsx` logic to:
   - Validate last visited board ID against current team or global scope
   - Iterate through categories to find the first visible, valid board
   - Fallback to first available view if last viewed ID is unavailable
   - Reduce duplicate API calls when switching teams

4. **Local Storage Cleanup** – Clear cached board/view IDs when deleting a board to prevent orphaned references

5. **Board Navigation** – Enhanced `showBoard` utility to save last viewed board in localStorage by team

---

## Change Impact: **High**

**Regression Risk:** HIGH – Changes modify core routing logic, error boundary behavior, and board/team validation across multiple layers. New validation paths may have untested edge cases, particularly around team switching and board resolution.

**High-Risk Flows Affected:** 
- Board loading and team switching
- Error recovery and error boundary redirect logic
- View/board selection when navigating to invalid routes
- First-time board creation for new users (reported issue being addressed)

**Blast Radius:** Wide – Affects all users navigating between boards and teams; touches shared navigation utilities, state management selectors, and error handling that impact fundamental application workflow.

**Reasoning:** This change addresses a critical user-facing bug (redirect loop) but introduces extensive validation logic across multiple core components (error boundary, board page routing, board/view selection). While these changes fix known issues, they substantially alter navigation behavior and error recovery paths, creating potential for new edge cases in untested scenarios (particularly noted: first-time board creation issues).

### DIFF
--- a/server/boards/boardsapp_util.go
+++ b/server/boards/boardsapp_util.go
@@ -80,7 +80,7 @@ func createBoardsConfig(mmconfig mm_model.Config, baseURL string, serverID strin
 		showFullName = *mmconfig.PrivacySettings.ShowFullName
 	}
 
-	serverRoot := baseURL + "/plugins/focalboard"
+	serverRoot := baseURL + "/boards"
 
 	return &config.Configuration{
 		ServerRoot:               serverRoot,
--- a/webapp/src/components/sidebar/sidebarCategory.tsx
+++ b/webapp/src/components/sidebar/sidebarCategory.tsx
@@ -179,6 +179,10 @@ const SidebarCategory = (props: Props) => {
             return
         }
         telemetryClient.trackEvent(TelemetryCategory, TelemetryActions.DeleteBoard, {board: deleteBoard.id})
+        
+        // Capture the category ID before deletion
+        const deletedFromCategoryID = props.categoryBoards.id
+        
         mutator.deleteBoard(
             deleteBoard,
             intl.formatMessage({id: 'Sidebar.delete-board', defaultMessage: 'Delete board'}),
@@ -197,10 +201,12 @@ const SidebarCategory = (props: Props) => {
                 }
             },
             async () => {
+                // Restore the board to the category it was deleted from
+                await mutator.moveBoardToCategory(teamID, deleteBoard.id, deletedFromCategoryID, '')
                 showBoard(deleteBoard.id)
             },
         )
-    }, [showBoard, deleteBoard, props.boards])
+    }, [showBoard, deleteBoard, props.boards, props.categoryBoards.id, teamID])
 
     const updateCategory = useCallback(async (value: boolean) => {
         const updatedCategory: Category = {
--- a/webapp/src/components/sidebar/sidebarCategory.tsx
+++ b/webapp/src/components/sidebar/sidebarCategory.tsx
@@ -39,6 +39,7 @@ import {TOUR_SIDEBAR, SidebarTourSteps, TOUR_BOARD, FINISHED} from '../../compon
 import telemetryClient, {TelemetryActions, TelemetryCategory} from '../../telemetry/telemetryClient'
 
 import {getCurrentTeam} from '../../store/teams'
+import {UserSettings} from '../../userSettings'
 
 import ConfirmationDialogBox, {ConfirmationDialogBoxProps} from '../confirmationDialogBox'
 
@@ -182,6 +183,10 @@ const SidebarCategory = (props: Props) => {
         
         // Capture the category ID before deletion
         const deletedFromCategoryID = props.categoryBoards.id
+        
+        // Clear localStorage entries for the deleted board
+        UserSettings.setLastBoardID(teamID, null)
+        UserSettings.setLastViewId(deleteBoard.id, null)
         
         mutator.deleteBoard(
             deleteBoard,
--- a/webapp/src/pages/boardPage/boardPage.tsx
+++ b/webapp/src/pages/boardPage/boardPage.tsx
@@ -53,7 +53,7 @@ import TelemetryClient, {TelemetryActions, TelemetryCategory} from '../../teleme
 
 import {Constants} from '../../constants'
 
-import {getCategoryOfBoard, getHiddenBoardIDs} from '../../store/sidebar'
+import {getCategoryOfBoard, getHiddenBoardIDs, removeBoardsFromAllCategories} from '../../store/sidebar'
 
 import SetWindowTitleAndIcon from './setWindowTitleAndIcon'
 import TeamToBoardAndViewRedirect from './teamToBoardAndViewRedirect'
@@ -138,6 +138,12 @@ const BoardPage = (props: Props): JSX.Element => {
                     teamId,
                     boardId: activeBoardId,
                 }))
+            }
+
+            // remove boards from all categories if they are deleted
+            const deletedBoardIds = teamBoards.filter((b: Board) => b.deleteAt && b.deleteAt !== 0).map((b) => b.id)
+            if (deletedBoardIds.length > 0) {
+                dispatch(removeBoardsFromAllCategories(deletedBoardIds))
             }
         }
 
--- a/webapp/src/pages/boardPage/boardPage.tsx
+++ b/webapp/src/pages/boardPage/boardPage.tsx
@@ -86,6 +86,7 @@ const BoardPage = (props: Props): JSX.Element => {
     const allTeams = useAppSelector(getAllTeams)
     const allBoards = useAppSelector(getBoards)
     const boardViews = useAppSelector(getCurrentBoardViews)
+    const [loadedBoardId, setLoadedBoardId] = useState<string>('')
 
     // if we're in a legacy route and not showing a shared board,
     // redirect to the new URL schema equivalent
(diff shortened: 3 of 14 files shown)


webapp/src/components/sidebar/sidebarCategory.tsx
```diff
@@ -183,6 +184,10 @@ const SidebarCategory = (props: Props) => {
         // Capture the category ID before deletion
         const deletedFromCategoryID = props.categoryBoards.id
         
+        // Clear localStorage entries for the deleted board
+        UserSettings.setLastBoardID(teamID, null)
```
Rajat-Dabade, line 188: More of the clean up: delete boardId and viewId from local storage when board is deleted

webapp/src/utils.ts
```diff
@@ -575,7 +576,8 @@ class Utils {
     }
 
     static getFrontendBaseURL(absolute?: boolean): string {
-        let frontendBaseURL = window.frontendBaseURL || Utils.getBaseURL()
+        // Always use /boards as the frontend base URL, never fall back to baseURL
+        let frontendBaseURL = window.frontendBaseURL || '/boards'
```
Rajat-Dabade: We do not use `plugins/focalboard` from frontend routes, so removed it.

devinbinnie (approved):

webapp/src/utils.ts
marianunez: It seems that `window.frontendBaseURL` is already set to `/boards here: 

https://github.com/mattermost/mattermost-plugin-boards/blob/56e47bac77e2af09e0bab55fcd68c0071a9ee358/webapp/src/index.tsx#L57

so this or seems redundant
```suggestion
        let frontendBaseURL = window.frontendBaseURL
```

marianunez (changes requested):
Filling in for QA Review: 

I am not able to reproduce the original redirect problem. However, testing out general links and redirects, found 2 regressions: 

1. When you link an invalid board (copied a valid board URL and changed a number), it no longer takes you to the error page but the Create Board page. The behavior should remain the same and be able to take you to a proper error page if it's not a valid board. 

2. When clicking the link the bot sends when a new board has been linked to a channel, it goes to a blank page. It seems that before it would resolve properly the `/plugins/focalboard/team/` paths. 

https://github.com/user-attachments/assets/6ccceac1-b6d4-4b0a-ac11-9e684ff95444

webapp/src/utils.ts
Rajat-Dabade: Added the ternary operator to make TypeScript happy.

marianunez (changes requested):
@Rajat-Dabade we need to make sure to validate the changes with more close attention as there was more routing changes introduced and now I'm seeing more regressions. 

> > When you link an invalid board (copied a valid board URL and changed a number), it no longer takes you to the error page but the Create Board page. The behavior should remain the same and be able to take you to a proper error page if it's not a valid board.
> 
> Fixed it ✅
> 

- It did indeed fix the issue of when navigating to an invalid path takes you to the error screen and even though when clicking `Back to Team` it restores you back to Boards, when you navigate back out to channels and back to Boards, it never recovers from that error state until you refresh. See video


https://github.com/user-attachments/assets/a0656cf6-0fc2-4b5f-b4ad-6641d3d22592

> > When clicking the link the bot sends when a new board has been linked to a channel, it goes to a blank page. It seems that before it would resolve properly the /plugins/focalboard/team/ paths.
> 
> This was the existing issue, nice catch. Fixed it ✅

I couldn't validate this one because now Creating a Board from a channel `Link Boards to Channel` is broken. It opens the create board screen with no template and none of the button work, even Create Empty Board does not work. See video


https://github.com/user-attachments/assets/46c98b61-1adb-498c-a5a0-3c958f67a6f7

server/boards/boardsapp_util.go
```diff
@@ -80,7 +80,7 @@ func createBoardsConfig(mmconfig mm_model.Config, baseURL string, serverID strin
 		showFullName = *mmconfig.PrivacySettings.ShowFullName
 	}
 
-	serverRoot := baseURL + "/plugins/focalboard"
+	serverRoot := baseURL + "/boards"
```
harshilsharma63, line 83: Are we sure this works? Wondering how this used to work before this change. And we should also verify this change in desktop app as IIRC, there is some boards and playbook specific routing code in the desktop app that is dependent on the URL.

webapp/src/components/sidebar/sidebarCategory.tsx
harshilsharma63, line 188: We should only do this if the boards being deleted is the current board. Currently if you are on Board A and Delete Board B, this clears the last board ID while in fact the last last board ID should continue pointing to Board A.

Additionally, I think this is causing one more issue where if you have Board A, B, C and D in that order, and you are on D and delete B, you end up navigating to C, but you should continue to stay on D

https://github.com/user-attachments/assets/5069b68f-4bfa-4b6a-b2ff-7dbc17591833

webapp/src/components/sidebar/sidebarCategory.tsx
```diff
@@ -183,6 +184,10 @@ const SidebarCategory = (props: Props) => {
...
+        UserSettings.setLastViewId(deleteBoard.id, null)
```
harshilsharma63, line 189: Same applied to this - only delete if deleting the current view.

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -102,10 +107,18 @@ const BoardPage = (props: Props): JSX.Element => {
...
+        if (allTeams.length === 0) {
+            return
+        }
+        const isValidTeam = allTeams.some((team) => team.id === match.params.teamId) || match.params.teamId === Constants.globalTeamId
+        if (isValidTeam) {
+            UserSettings.lastTeamId = teamId
+            octoClient.teamId = teamId
+            dispatch(setTeam(teamId))
+        } else {
+            dispatch(setGlobalError('team-undefined'))
```
harshilsharma63, line 119: I'm not sure what this does, but for example if I open the url `http://localhost:8065/boards/team/invalid-team-id` on this branch, there are no errors and an empty template selector screen-

Image

Whereas if I try to do the same thing on Hub, I get the screen with right error message-

Image

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -229,28 +257,62 @@ const BoardPage = (props: Props): JSX.Element => {
...
+                UserSettings.setLastViewId(match.params.boardId, viewId)
             }
         }
-    }, [teamId, match.params.boardId, viewId, me?.id])
 
+        dispatch(loadAction(match.params.boardId))
+    }, [teamId, match.params.boardId, viewId, me?.id, dispatch, loadAction])
+
+
+    useEffect(() => {
```
harshilsharma63, line 276: This creates a new bug. If you have two teams with following boards-

1. Team A - Board A, B and C
2. Team B - Board D, E, F

If you open team B in boards and open board E, navigate back to Channels, then back to boards, it should stay you on Board E, but currently it navigates to first board D.

Additionally, if you open Team B Board E, then switch to team A, then to team B, it ends up switching to board D but it should stay on the board which was previously open.

I'm not sure whats the need of resetting board ID on team switch when we store each team's last board ID in local storage.

harshilsharma63 (changes requested):

marianunez:
@coderabbitai please review

marianunez:
@coderabbitai do a full review and include the summary in the PR description

marianunez:
Fixed with https://github.com/mattermost/mattermost-plugin-boards/pull/140

PR #141, opened 2025-10-10
Fix the reconnect issue for boards

Rajat-Dabade:
#### Summary

I was not able to reproduce this locally, but suspect this could happen when the board is stale for a long period. 

When there is no activity on Mattermost (like on the weekend), the websocket becomes stale. And when the boards tab is open, the following could happen:
- Long inactivity, stale websocket and session expired. 
- API calls fail due to a stale session. 
- Error is thrown in `initialLoad`
- Global error state set. 
- Redirected to Error page.
- "Sorry, Something went wrong" page is displayed. 
- And polling starts infinitely to reconnect. 
- After the hard refresh, everything fixes. 

#### Solution: 
- Added registry-based reconnection handler that triggers initialLoad() when WebSocket reconnects (same pattern as Playbooks).
- Removed infinite polling mechanism when the connection closes.

#### How this will work:
- User returns after the weekend (or after a longer time).
- User clicks on the Boards tab. 
- WebSocket closes → Mattermost automatically handles reconnection
- Registry handler triggers initialLoad() on reconnect
- Board load successfully. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63248

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -62,12 +62,6 @@ func NewAPI(
 }
 
 func (a *API) RegisterRoutes(r *mux.Router) {
-	// V1 routes for simple endpoints
-	apiv1 := r.PathPrefix("/api/v1").Subrouter()
-	apiv1.Use(a.panicHandler)
-	apiv1.Use(a.requireCSRFToken)
-	a.registerKeepaliveRoute(apiv1)
-
 	apiv2 := r.PathPrefix("/api/v2").Subrouter()
 	apiv2.Use(a.panicHandler)
 	apiv2.Use(a.requireCSRFToken)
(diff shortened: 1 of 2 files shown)


server/api/api.go
```diff
@@ -62,6 +62,12 @@ func NewAPI(
 }
 
 func (a *API) RegisterRoutes(r *mux.Router) {
+	// V1 routes for simple endpoints
+	apiv1 := r.PathPrefix("/api/v1").Subrouter()
+	apiv1.Use(a.panicHandler)
+	apiv1.Use(a.requireCSRFToken)
+	a.registerKeepaliveRoute(apiv1)
```
marianunez: Why are we adding back a v1 route? I'm guessing this version was existing in the first versions of Focalboard and were deprecated in favor for v2.

webapp/src/index.tsx
```diff
@@ -383,6 +394,36 @@ export default class Plugin {
...
+
+        this.activityFunc = () => {
+            const now = new Date().getTime()
+            if (now - lastActivityTime > activityTimeout) {
+                this.notifyConnect()
+            }
+            lastActivityTime = now
+        }
+        document.addEventListener('click', this.activityFunc)
+    }
```
marianunez: Why do we need to keep track of user activity here to keep websocket connection alive? If the user is interacting with the app it will naturally keep the websockets active

server/api/api.go
Rajat-Dabade: Ah, yes. That's the nice catch. Will update the code.

webapp/src/index.tsx
Rajat-Dabade: Now that I think about this one, we already have the registerReconnectHandler, which will be fired when we reconnect. So I think we do not require the keep-alive API call. Will remove it.

PR #146, opened 2025-11-03
Fix: Performing REDO on attachment bring the attachment back

Rajat-Dabade:
#### Summary
- User uploads a file attachment to a card
  - File stored with DeleteAt = 0 (active)
  - Block created with fileId pointing to the file
- User presses Undo (`Cmd+Z`)
  - Block is soft-deleted (DeleteAt set to current timestamp)
  - File is also soft-deleted (FileInfo.DeleteAt set to current timestamp)
  - This worked fine
- User presses Redo (`Cmd+Shift+Z`) THIS BROKE
  - Frontend creates a NEW block (with a new ID) instead of undeleting the old one
  - New block is inserted with the same fileId reference
  - BUT: The file's DeleteAt was never reset to 0
  - Result: Block exists, but the file is still marked as deleted
  - When trying to access the file → 403 Forbidden / 404 Not Found

This PR added the file restoration logic when inserting blocks. This will check if a block has an image or attachments and restore it. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-66435

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,3 +42,10 @@ jobs:
     
     - name: Lint & test server 
       run: cd focalboard; make server-ci
+      
+  dist:
+    uses: mattermost/actions-workflows/.github/workflows/plugin-dist-pr.yml@main
+    secrets: inherit
+    with:
+      dist-target: "dist-linux"
+      s3-prefix: "mattermost-plugin-boards"
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: prebuild clean cleanall ci server server-mac server-linux server-win server-linux-package generate watch-server webapp mac-app win-app-wpf linux-app modd-precheck templates-archive
+.PHONY: prebuild clean cleanall ci server server-linux server-mac server-win server-linux-package generate watch-server webapp mac-app win-app-wpf linux-app modd-precheck templates-archive
 
 PACKAGE_FOLDER = focalboard
 
@@ -114,6 +114,20 @@ else
 endif
 endif
 
+## Builds the server, if it exists, for Linux architectures only.
+.PHONY: server-linux
+server-linux: templates-archive
+ifneq ($(HAS_SERVER),)
+	mkdir -p server/dist;
+ifeq ($(MM_DEBUG),)
+	cd server && env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build $(GO_BUILD_FLAGS) -trimpath -o dist/plugin-linux-amd64;
+else
+	$(info DEBUG mode is on; to disable, unset MM_DEBUG)
+
+	cd server && env CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build $(GO_BUILD_FLAGS) -gcflags "all=-N -l" -trimpath -o dist/plugin-linux-amd64;
+endif
+endif
+
 ## Ensures NPM dependencies are installed without having to run this all the time.
 webapp/node_modules: $(wildcard webapp/package.json)
 ifneq ($(HAS_WEBAPP),)
@@ -175,6 +189,10 @@ info: ## Display build information
 ## Builds and bundles the plugin.
 .PHONY: dist
 dist:	apply server webapp bundle
+
+## Builds and bundles the plugin for Linux only.
+.PHONY: dist-linux
+dist-linux:	apply server-linux webapp bundle
 
 ## Builds and installs the plugin to a server.
 .PHONY: deploy
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Mattermost Boards
 
-### A self-hosted plugin for managing boards withing a Mattermost installation
+### A self-hosted plugin for managing boards within a Mattermost installation
 
 Mattermost boards plugins is an open source, multilingual, self-hosted project management tool that's an alternative to Trello, Notion, and Asana.
 
--- a/webapp/src/mutator.ts
+++ b/webapp/src/mutator.ts
@@ -1050,8 +1050,9 @@ class Mutator {
                 }
                 return [blocks, newRootBlock.id]
             },
-            async (newBlocks: Block[]) => {
+            async (result: [Block[], string]) => {
                 await beforeUndo?.()
+                const [newBlocks] = result
                 const newRootBlock = newBlocks && newBlocks[0]
                 if (newRootBlock) {
                     await octoClient.deleteBlock(newRootBlock.boardId, newRootBlock.id)


server/services/store/sqlstore/blocks.go
```diff
@@ -245,6 +245,37 @@ func (s *SQLStore) insertBlock(db sq.BaseRunner, block *model.Block, userID stri
...
+				fileID := retrieveFileIDFromBlockFieldStorage(attachmentIDStr)
+				if fileID != "" {
+					fileIDsToRestore = append(fileIDsToRestore, fileID)
+				}
+			}
+		}
+
+		if len(fileIDsToRestore) > 0 {
+			if err := s.restoreFiles(db, fileIDsToRestore); err != nil {
+				s.logger.Error(
```
pvev: question: should this error be propagated to webapp? like how'd be the UX on this scenario?

server/services/store/sqlstore/blocks.go
```diff
@@ -245,6 +245,37 @@ func (s *SQLStore) insertBlock(db sq.BaseRunner, block *model.Block, userID stri
 		return err
 	}
 
+	if existingBlock == nil && (block.Type == "image" || block.Type == "attachment") {
+		fileIDsToRestore := make([]string, 0, 2)
```
pvev: question: do you expect fileId and attachmentId to exist at the same time and that's why the 2 capaciy?

server/services/store/sqlstore/blocks.go
Rajat-Dabade: Great point. This is primarily done for the redo operation. Even if the file restoration fails, the user can always re-upload the file if needed. I agree we need a better UX and maybe in future we can add it. But for now I believe failing siliently will not cause any major issue.

server/services/store/sqlstore/blocks.go
Rajat-Dabade: The legacy code was using `attachmentId` to store attachments, but it was then migrated to use `fileId`. That's the reason why 2 checks are needed here.

server/services/store/sqlstore/blocks.go
```diff
@@ -245,6 +245,37 @@ func (s *SQLStore) insertBlock(db sq.BaseRunner, block *model.Block, userID stri
 		return err
 	}
 
+	if existingBlock == nil && (block.Type == "image" || block.Type == "attachment") {
```
harshilsharma63: All this logic should be in the App layer, not the store layer. Store layer shouldn't be concerned with this.

server/services/store/sqlstore/blocks.go
```diff
@@ -245,6 +245,37 @@ func (s *SQLStore) insertBlock(db sq.BaseRunner, block *model.Block, userID stri
...
+		if fileIDVal, exists := block.Fields["fileId"]; exists {
+			if fileIDStr, ok := fileIDVal.(string); ok && fileIDStr != "" {
+				fileID := retrieveFileIDFromBlockFieldStorage(fileIDStr)
+				if fileID != "" {
+					fileIDsToRestore = append(fileIDsToRestore, fileID)
+				}
+			}
+		}
+
+		if attachmentIDVal, exists := block.Fields["attachmentId"]; exists {
```
harshilsharma63: Since we migrated away from `attachmentId`, we should no longer have any code to continue supporting it. If we still think some users have `attachmentId`, we can add a migration to update it. Otherwise we're basically supporting two ways of doing the same thing and we'd have to always keep supporting both keys.

harshilsharma63 (changes requested):

server/services/store/sqlstore/blocks.go
Rajat-Dabade: Okay.

server/services/store/sqlstore/blocks.go
Rajat-Dabade: This is a good point. Will definitely create a ticket as it is out of scope for this PR.

Rajat-Dabade: https://mattermost.atlassian.net/browse/MM-66557

server/utils/utils.go
```diff
@@ -128,3 +129,13 @@ func DedupeStringArr(arr []string) []string {
 func GetBaseFilePath() string {
 	return path.Join("boards", time.Now().Format("20060102"))
 }
+
+// RetrieveFileIDFromBlockFieldStorage extracts the file ID from the stored format.
+// The stored format is typically "{fileID}.{extension}", so we split by "." and extract the fileID.
```
enzowritescode: Saying that is "typically" in that format makes it seem like we are not confident that this always the case. I just want to sanity check on that. Either we should remove "typically" or provide other examples and some accompanying tests.

server/app/blocks.go
```diff
@@ -193,13 +194,34 @@ func (a *App) InsertBlocksAndNotify(blocks []*model.Block, modifiedByID string,
...
+
+		existingBlock, checkErr := a.store.GetBlock(block.ID)
+		if checkErr != nil && !model.IsErrNotFound(checkErr) {
+			return nil, checkErr
+		}
+
+		if existingBlock == nil && (block.Type == "image" || block.Type == "attachment") {
+			fileIDsToRestore := extractFileIDsFromBlock(block)
+			if len(fileIDsToRestore) > 0 {
+				if restoreErr := a.store.RestoreFiles(fileIDsToRestore); restoreErr != nil {
```
enzowritescode: Missing authorization check. User A creates and deletes an image on Board A. User B (who doesn't have access to Board A) can then restore it if they know the fileId by creating a block on Board B with the same fileId. I couldn't find what side effect this had in the UI, but regardless User B should not be able to perform actions on things they don't have access to.

User A creates image:
Image

User A deletes image:
Image

User B restores image:
Image

server/app/blocks.go
```diff
@@ -387,3 +409,28 @@ func (a *App) getBoardAndCard(block *model.Block) (board *model.Board, card *mod
...
+	if fileIDVal, exists := block.Fields["fileId"]; exists {
+		if fileIDStr, ok := fileIDVal.(string); ok && fileIDStr != "" {
+			fileID := utils.RetrieveFileIDFromBlockFieldStorage(fileIDStr)
+			if fileID != "" {
+				fileIDsToRestore = append(fileIDsToRestore, fileID)
+			}
+		}
+	}
+
+	if attachmentIDVal, exists := block.Fields["attachmentId"]; exists {
```
harshilsharma63: Why can't we remove the attachmentId field right now? I mean, its just about removing this code that checks for attachmentId field.

To clarify, I'm not asking to remove all usages of attachmentId from the whole codebase in this PR, just to not add any new code that references that field.

Rajat-Dabade: Okay.

server/utils/utils.go
Rajat-Dabade: This is always the case, will remove the typically.

server/app/blocks.go
Rajat-Dabade: That's a great catch! Even if the file gets restored, User B still wouldn’t be able to access it since we already have a check in place that blocks non-members from viewing board data. But you’re right, we should also prevent unauthorised restoration of files. I will add a check for that as well.

harshilsharma63 (approved):

PR #152, opened 2025-12-16
Validation to modify comments

Rajat-Dabade:
#### Summary
This PR adds a check that prevents other users from modifying others' comments. Only the comment creator or user with `PermissionDeleteOthersComments` can modify comments. Also fixed the `insertBlock` function to preserve the original creator in the `blocks_history` table. Previously, the history incorrectly showed the modifier as the creator. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66752

### DIFF
--- a/server/api/blocks.go
+++ b/server/api/blocks.go
@@ -570,6 +570,13 @@ func (a *API) handlePatchBlock(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if block.Type == model.TypeComment && block.CreatedBy != userID {
+		if !a.permissions.HasPermissionToBoard(userID, boardID, model.PermissionDeleteOthersComments) {
+			a.errorResponse(w, r, model.NewErrPermission("access denied to modify others' comments"))
+			return
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


server/api/blocks.go
```diff
@@ -570,6 +570,13 @@ func (a *API) handlePatchBlock(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if block.Type == model.TypeComment && block.CreatedBy != userID {
```
harshilsharma63, line 573: While the change looks good for this fix, is there any case where a user is allowed to modify someone else's block? As in, should the condition `block.Type == model.TypeComment` even exist?

harshilsharma63 (approved):

server/api/blocks.go
harshilsharma63: Does similar check exist in all of CRUD APIs for boards and blocks?

Rajat-Dabade:
@pvev, the server tests are disabled on the CI as of now. I am working on fixing them in a separate PR. Will add the test for this one in that PR itself.

PR #153, opened 2025-12-30
Update mattermost/server/public to v0.1.21

esarafianou:
#### Summary 
* Updates mattermost/server/public to v0.1.21
* Removes mySQL related code. Updating the server package resulted in errors due to Boards requiring SQL util functions that were no longer available in 0.1.21

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-67058

### DIFF
--- a/server/services/store/sqlstore/migrationstests/boards_migrator_test.go
+++ b/server/services/store/sqlstore/migrationstests/boards_migrator_test.go
@@ -14,13 +14,11 @@ import (
 	"github.com/mattermost/mattermost/server/public/pluginapi/cluster"
 	"github.com/mattermost/morph"
 	"github.com/mattermost/morph/drivers"
-	"github.com/mattermost/morph/drivers/mysql"
 	"github.com/mattermost/morph/drivers/postgres"
 	"github.com/mattermost/morph/drivers/sqlite"
 	embedded "github.com/mattermost/morph/sources/embedded"
 
 	"github.com/mattermost/mattermost/server/public/shared/mlog"
-	mmSqlStore "github.com/mattermost/mattermost/server/public/utils/sql"
 	"github.com/mattermost/mattermost/server/v8/channels/db"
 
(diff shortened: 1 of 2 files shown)


server/services/store/sqlstore/migrationstests/boards_migrator_test.go
```diff
@@ -94,11 +92,6 @@ func (bm *BoardsMigrator) getDriver() (drivers.Driver, error) {
 		if err != nil {
 			return nil, err
 		}
-	case model.MysqlDBType:
-		driver, err = mysql.WithInstance(bm.db)
-		if err != nil {
-			return nil, err
-		}
```
marianunez, line 101: @esarafianou it may be worth to move some of these utils definitions over to the plugin to not remove mySQL support as we still need to support v10.11 ESR that customers may still be running in mySQL. 

I would like to avoid having to manage release branches to update customers in pre-v11 versions with other security fixes given that we've dropped mySQL support early on the plugin,

marianunez (changes requested):

server/services/store/sqlstore/migrationstests/boards_migrator_test.go
esarafianou: Yep, that's sound like a much better approach than removing the mySQL support. Could the Integrations team continue the remaining work?

marianunez: Yes, @Rajat-Dabade are you able to make these changes to complete this dependency upgrade without removing mySQL support?

Rajat-Dabade: @marianunez  @esarafianou, this is taken care of as a part of this PR: https://github.com/mattermost/mattermost-plugin-boards/pull/158/changes

esarafianou: Thanks! Closing this one

PR #164, opened 2026-02-02
Instead of manually updating the version, reason from plugin.json

Rajat-Dabade:
#### Summary
Removed the need to manually update the board version in different parts of the code. Now, the version is automatically read from the plugin.json file and used everywhere.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67360

## Summary by CodeRabbit

* **Chores**
  * Updated build process to automatically derive plugin version and release notes URL from Git tags and build metadata.
  * Simplified plugin manifest by removing static version and release notes URL fields; these values are now generated dynamically at build time.


harshilsharma63:
We'd still need to make the other changes, for example, from this PR - https://github.com/mattermost/mattermost-plugin-boards/pull/156/changes right?

webapp/src/manifest.ts
```diff
@@ -1,7 +1,49 @@
 // Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import manifest from '../../plugin.json'
+// This file is automatically generated. Do not modify it manually.
+
+const manifest = JSON.parse(`
```
harshilsharma63, line 9: Not sure whats the reason behind this change. Can you add code comment explaining why we cannot import `plugin.json` as I think someone will be confused in the future as well.

Rajat-Dabade: Added the comments.

harshilsharma63: Ah, got it. It's an autogenerated file.

harshilsharma63 (approved):

PR #165, opened 2026-02-03
Board blank page if no team id is present in localstorage

Rajat-Dabade:
#### Summary
Found an issue where if no team Id is present in the localstorage, the board on refresh on the page with URL `boards/team` goes blank. This PR solves this issue. It fetches the team ID from the API call, looks for the first team and redirects it to that team. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67431

### DIFF
--- a/webapp/src/router.tsx
+++ b/webapp/src/router.tsx
@@ -70,13 +70,14 @@ function HomeToCurrentTeam(props: {path: string, exact: boolean}) {
 
                 let teamID = (window.getCurrentTeamId && window.getCurrentTeamId()) || ''
                 const lastTeamID = UserSettings.lastTeamId
+                const hasNoTeamId = !teamID && !firstTeam && !lastTeamID
                 
-                if (teamsFetchError && !teamID && !firstTeam && !lastTeamID) {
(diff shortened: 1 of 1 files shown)


webapp/src/router.tsx
```diff
@@ -38,26 +38,62 @@ function HomeToCurrentTeam(props: {path: string, exact: boolean}) {
...
                 let teamID = (window.getCurrentTeamId && window.getCurrentTeamId()) || ''
                 const lastTeamID = UserSettings.lastTeamId
-                if (!teamID && !firstTeam && !lastTeamID) {
+                
+                if (teamsFetchError && !teamID && !firstTeam && !lastTeamID) {
+                    history.replace('/error?id=unknown')
+                    return null
+                }
+                
+                if (!teamID && !firstTeam && !lastTeamID && !teamsFetched) {
```
devinbinnie: Nit: We're checking 3 of these conditions twice in a row. Could simplify this.

harshilsharma63 (approved):

webapp/src/router.tsx
Rajat-Dabade: Done.

PR #170, opened 2026-02-09
Use webapp exported markdown to card and board descriptions

Rajat-Dabade:
#### Summary
Now the card and board description uses webapp-exported markdown. 

https://github.com/user-attachments/assets/60a29110-faca-4ca3-be26-2bbc109424c0

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62511

### DIFF
--- a/webapp/src/components/markdownEditor.tsx
+++ b/webapp/src/components/markdownEditor.tsx
@@ -1,8 +1,9 @@
 // Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useState, Suspense} from 'react'
+import React, {useState, Suspense, useMemo} from 'react'
 
+import {Channel} from '@mattermost/types/channels'
 import {getChannelsNameMapInTeam} from 'mattermost-redux/selectors/entities/channels'
 
 import {Provider} from 'react-redux'
@@ -13,6 +14,8 @@ import {useAppSelector} from '../store/hooks'
 import './markdownEditor.scss'
 
 const MarkdownEditorInput = React.lazy(() => import('./markdownEditorInput/markdownEditorInput'))
+
+const EMPTY_CHANNEL_NAMES_MAP: Record<string, Channel> = {}
 
 type Props = {
     id?: string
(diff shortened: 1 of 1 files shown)


webapp/src/components/markdownEditor.tsx
```diff
@@ -27,13 +33,14 @@ type Props = {
 const MarkdownEditor = (props: Props): JSX.Element => {
     const {placeholderText, onFocus, onEditorCancel, onBlur, onChange, text, id, saveOnEnter} = props
     const [isEditing, setIsEditing] = useState(Boolean(props.autofocus))
-    const html: string = Utils.htmlFromMarkdown(text || placeholderText || '')
+
+    const selectedTeam = useAppSelector(getCurrentTeam)
+    const channelNamesMap = selectedTeam ? getChannelsNameMapInTeam((window as any).store.getState(), selectedTeam.id) : {}
```
devinbinnie: Is `window as any` something used in the app a lot? If so - might be worth typing the `Window` object for your app (not for this PR)
If not - can we change this or add the typing if it already exists?

webapp/src/components/markdownEditor.tsx
```diff
@@ -46,7 +53,29 @@ const MarkdownEditor = (props: Props): JSX.Element => {
                     setIsEditing(true)
                 }
             }}
-        />
+        >
+            {text ? (
```
harshilsharma63: The only difference between the two is `text` vs `placeholderText || ''`, everything else is the same. We can simplify this to `formatText(text || placeholderText || '')` and avoid duplication.

webapp/src/components/markdownEditor.tsx
harshilsharma63: We need to memomize this as otherwise this will be computed on every render cycle and possibly trigger rerenders of child components.

harshilsharma63 (approved):

webapp/src/components/markdownEditor.tsx
Rajat-Dabade: Yes, there are many places where `window as any` is used. I agree with the suggestion that we should type it. I will create a ticket and do this separately.

webapp/src/components/markdownEditor.tsx
Rajat-Dabade: Nice catch, will update it.

webapp/src/components/markdownEditor.tsx
Rajat-Dabade: Done.

yasserfaraazkhan (approved):

PR #176, opened 2026-03-11
Fix: Board infinity redirect multiperson property

Rajat-Dabade:
#### Summary

Root Cause:
The crash happened in sortCards in `webapp/src/store/cards.ts` when:
- The board was sorted by a multi-person property.
- The page was refreshed.
- `boardUsers` was still empty because `fetchBoardMembers` had not finished.
- For multiPerson, the code only converted arrays to strings when usersById was non-empty.
- When `usersById` was empty, `aValue` and `bValue` stayed as arrays (e.g. ["user_id_1", "user_id_2"]).
- The code then called `(aValue as string).localeCompare(bValue as string)`.
- At runtime, aValue was still an array, so localeCompare was called on an array and caused TypeError: `aValue.localeCompare` is not a function.
- That error was caught by the React ErrorBoundary, which redirected to `/error?id=unknown`.

Steps to reproduce: 
- Create a new board with the template "Company Goals and OKRs"
- Create a new "board view" in the top left
- On one of the existing cards in the category, add a multi-person property
- Add yourself to this field on two of the cards in the view
- Set the sort for the view to that multi-person property
- After this, clicking the 3 dots and then “copy link” should produce links with the bug for any card on that view. Changing the sort to anything other than that property should also fix it.

Before: 

https://github.com/user-attachments/assets/9aea7580-9212-4d8f-8dcc-5f611c2e2554

After: 

https://github.com/user-attachments/assets/59d50ca2-c76c-4bb9-b701-85213059a39e

#### Ticket Link

### DIFF
--- a/webapp/src/store/cards.ts
+++ b/webapp/src/store/cards.ts
@@ -257,8 +257,16 @@ function sortCards(cards: Card[], lastCommentByCard: {[key: string]: CommentBloc
                     aValue = usersById[a.modifiedBy]?.username || ''
                     bValue = usersById[b.modifiedBy]?.username || ''
                 } else if (template.type === 'date') {
-                    aValue = (aValue === '') ? '' : JSON.parse(aValue as string).from
-                    bValue = (bValue === '') ? '' : JSON.parse(bValue as string).from
+                    try {
+                        aValue = (aValue === '') ? '' : JSON.parse(aValue as string).from
+                    } catch {
+                        aValue = ''
+                    }
+                    try {
+                        bValue = (bValue === '') ? '' : JSON.parse(bValue as string).from
+                    } catch {
+                        bValue = ''
+                    }
                 }
 
                 let result = 0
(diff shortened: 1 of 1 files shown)


devinbinnie (approved):

marianunez:
> * That error was caught by the React ErrorBoundary, which redirected to `/error?id=unknown`.

@Rajat-Dabade As we discussed offline, it seems this PR fixes the issue that was causing the error. 

However, do we know why the redirects are happening when it finds an error? The customer reported that it kept looping an adding `/error` to the url like: `http://localhost:8065/boards/team/ragwjr9o3jgr8cep5zh3hqh5na/bhcnsg4cu3jyy5qdczeknsdwdfe/vq8ymopyhipnffjoi119dpyybra/boards/boards/boards/boards/boards/boards/boards/boards/error` - see the message [here](https://hub.mattermost.com/pde/pl/7jo4mzcfdjf58rqtjoqbkbu9hr)

We can complete this PR and ship this fix to the customer but it's important that we find the cause of the redirects when there are errors because there may be other instances that there are errors and it should be able to manage them gracefully.

marianunez:
Harshil is OOO so you probably need to swap him out for another reviewer

Rajat-Dabade:
@marianunez Sure, I was finding that PR, but unable to find the fix in that PR.  I will try to see if I can do this in the same PR itself.

webapp/src/store/cards.ts
```diff
@@ -300,19 +308,29 @@ function sortCards(cards: Card[], lastCommentByCard: {[key: string]: CommentBloc
...
-                            return ''
-                        }).toString() : aValue
-
-                        bValue = Array.isArray(bValue) && bValue.length !== 0 && Object.keys(usersById).length > 0 ? bValue.map((id) => {
-                            if (usersById[id] !== undefined) {
-                                return usersById[id].username
-                            }
-                            return ''
-                        }).toString() : bValue
+                        if (Array.isArray(aValue) && aValue.length !== 0 && Object.keys(usersById).length > 0) {
```
larkox: I feel we would have the same result if we just skip the two conditions here (and in the next if) and remove the else.

If `aValue` is an empty array, we are returning an empty string either way. So not big deal.

The only case where we would be returning something different (and not sure what is the expected behavior) would be if the `usersById` is empty. In that scenario, the old implementation would break, this implementation would return `'id1,id2,id3`, and my proposal would return `,,` (more in line with what would happen if a user doesn't exist (`userA,,userC`).

Rajat-Dabade: Yes, that makes sense. The previous implementation was sorting by ID, so I initially kept it the same. But now I realise that doesn’t really make sense,  it just adds unnecessary complexity, and the sorting looks random, which can also make the UI jumpy.

I understand your point now: since we don’t know the username yet, it is better to treat all unresolved users equally instead of sorting them based on an ID that doesn’t really mean anything.

larkox (approved):

webapp/src/store/cards.ts
larkox: I don't understand the reasoning about the sorting, but looks good to me now.

larkox:
Does this need any kind of test?

Rajat-Dabade:
> Does this need any kind of test?

CI is running no tests as of now. We are working on enabling it. Will cover the test in that PR.

Rajat-Dabade:
@yasserfaraazkhan, I have tested this one, and it seems to be working fine on the Spinwick server. You can verify it once. Thank you.

yasserfaraazkhan (approved):

PR #155, opened 2026-01-08
Fix Server test for boards

Rajat-Dabade:
#### Summary
Fixing all the server tests for boards.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67641

## Change Impact: 🔴 **High**

**Regression Risk:** The PR introduces several significant regression risks:

1. **Removed Public Endpoint**: The `/teams/{teamID}/archive/export` endpoint is completely removed from `server/api/teams.go` and `server/api/archive.go`, but the integration tests in `permissions_test.go` still contain ~30 test cases explicitly validating this endpoint's behavior across different user roles and board types. These tests are now orphaned and will fail, indicating incomplete cleanup or broken test suite.

2. **Authorization Flow Refactoring**: Comprehensive replacement of session-context-derived user lookup with `getUserID(r)` across 8+ API handler files (archive.go, blocks.go, boards.go, categories.go, members.go, subscriptions.go, teams.go, users.go, content_blocks.go). This touches core permission enforcement paths with potential for authorization bypass or false denials if `getUserID()` behaves differently under edge cases.

3. **Inconsistent Database Driver Support**: The PR claims to remove MySQL support in team.go (using PostgreSQL-only `ON CONFLICT` syntax), but 6 other sqlstore files (`system.go`, `subscriptions.go`, `sharing.go`, `notificationhints.go`, `category_boards.go`, `board.go`) still retain MySQL-specific `ON DUPLICATE KEY UPDATE` syntax. This inconsistency creates undefined behavior for MySQL deployments.

4. **Complete Test Infrastructure Migration**: All 179 integration test calls migrated from token-based `SetupTestHelper().InitBasic()` to plugin-mode `SetupTestHelperPluginMode()` with zero fallback paths. If plugin-mode infrastructure has issues, entire test suite fails with no regression detection.

5. **Test Deletions**: Removal of entire test files (`localpermissions_test.go` - 172 lines, `session.go` tests - 37 lines) reduces regression detection coverage on permission and session handling logic.

**QA Recommendation:** **Mandatory comprehensive manual QA** in addition to automated testing:
- **Critical**: Verify whether the orphaned `/teams/{teamID}/archive/export` endpoint tests in `permissions_test.go` are intentionally disabled or represent a test suite break. If tests are failing, entire QA sign-off is unreliable.
- **Authorization paths**: All board/block/team operations across guest users, template access, and role-based views (Viewer, Commenter, Editor) must be manually validated for both permission grants and denials.
[...]

PR #177, opened 2026-03-17
Fix the infinite redirect loop for boards

Rajat-Dabade:
#### Summary
The PR Fix the infinite redirect loop for boards. 

### Findings on the Infinite Redirect Issue for Boards

While trying to reproduce the infinite redirect issue in Boards, I noticed a pattern related to URLs that end with `/error`.
When the page is reloaded while the URL contains `/error`, the following sequence happens:

- Initial URL
Example URL:
```
http://localhost:8065/boards/team/<teamId>/<boardId>/<viewId>/error
```

- Error boundary triggers
The application error boundary catches an error and runs this redirect:
```
window.location.replace('plugins/focalboard/error?id=unknown')
```

- Browser treats the redirect as a relative path
Since the path doesn’t start with /, the browser appends it to the current URL instead of treating it as a root path.
So the new URL becomes:
```
http://localhost:8065/boards/team/<teamId>/<boardId>/<viewId>/plugins/focalboard/error?id=unknown
```

- Route mismatch happens
This new URL does not match the `/error` route, so the app loads BoardPage (or a similar route) with incorrect parameters.

-  Another error is triggered
BoardPage fails again, which causes:
  - another error
  - the error boundary to run again
  - another redirect

-  URL keeps growing
Each redirect appends more segments to the path, such as:
```
.../plugins/focalboard/error
.../boards/error
.../boards/plugins/focalboard/error
```
This creates the infinite redirect loop.

#### Why the error boundary kept triggering
Routes like:
```
/boards/team/.../error
```

are matched by the BoardPage route:
```
/team/:teamId/:boardId?/:viewId?/:cardId?
```
In this case: cardId = "error"

BoardPage then tries to load a card with ID “error”, which does not exist.
This throws an error, which the error boundary catches again, restarting the same redirect process.

#### Ticket Link

### DIFF
--- a/webapp/src/error_boundary.tsx
+++ b/webapp/src/error_boundary.tsx
@@ -5,6 +5,7 @@
 import React from 'react'
 
 import {Utils} from './utils'
+import {Constants} from './constants'
 
 type State = {
     hasError: boolean
@@ -19,8 +20,33 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     msg = 'Redirecting to error page...'
 
     handleError = (): void => {
-        const url = Utils.getBaseURL() + '/error?id=unknown'
-        Utils.log('error boundary redirecting to ' + url)
+        if (window.location.pathname.endsWith('/error')) {
+            return
+        }
+
+        // Loop detection: prevent infinite error redirects
+        const now = Date.now()
+        const lastRedirectTime = parseInt(sessionStorage.getItem(Constants.sessionStorageErrorRedirectTimeKey) || '0', 10)
+        const redirectCount = parseInt(sessionStorage.getItem(Constants.sessionStorageErrorRedirectCountKey) || '0', 10)
+
+        if (now - lastRedirectTime > 10000) {
+            sessionStorage.setItem(Constants.sessionStorageErrorRedirectCountKey, '1')
+            sessionStorage.setItem(Constants.sessionStorageErrorRedirectTimeKey, now.toString())
+        } else {
+            const newCount = redirectCount + 1
+            sessionStorage.setItem(Constants.sessionStorageErrorRedirectCountKey, newCount.toString())
+            sessionStorage.setItem(Constants.sessionStorageErrorRedirectTimeKey, now.toString())
+
+            // If redirected more than 3 times in 10 seconds, go to boards root
+            if (newCount > 3) {
+                sessionStorage.removeItem(Constants.sessionStorageErrorRedirectCountKey)
+                sessionStorage.removeItem(Constants.sessionStorageErrorRedirectTimeKey)
+                window.location.replace(Utils.getFrontendBaseURL(true))
+                return
+            }
+        }
+
+        const url = Utils.getFrontendBaseURL(true) + '/error?id=unknown'
         window.location.replace(url)
     }
 
--- a/webapp/src/error_boundary.tsx
+++ b/webapp/src/error_boundary.tsx
@@ -5,7 +5,6 @@
 import React from 'react'
 
 import {Utils} from './utils'
-import {Constants} from './constants'
 
 type State = {
     hasError: boolean
@@ -22,28 +21,6 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     handleError = (): void => {
         if (window.location.pathname.endsWith('/error')) {
             return
-        }
-
-        // Loop detection: prevent infinite error redirects
-        const now = Date.now()
-        const lastRedirectTime = parseInt(sessionStorage.getItem(Constants.sessionStorageErrorRedirectTimeKey) || '0', 10)
-        const redirectCount = parseInt(sessionStorage.getItem(Constants.sessionStorageErrorRedirectCountKey) || '0', 10)
-
-        if (now - lastRedirectTime > 10000) {
-            sessionStorage.setItem(Constants.sessionStorageErrorRedirectCountKey, '1')
-            sessionStorage.setItem(Constants.sessionStorageErrorRedirectTimeKey, now.toString())
-        } else {
-            const newCount = redirectCount + 1
-            sessionStorage.setItem(Constants.sessionStorageErrorRedirectCountKey, newCount.toString())
-            sessionStorage.setItem(Constants.sessionStorageErrorRedirectTimeKey, now.toString())
-
-            // If redirected more than 3 times in 10 seconds, go to boards root
-            if (newCount > 3) {
-                sessionStorage.removeItem(Constants.sessionStorageErrorRedirectCountKey)
-                sessionStorage.removeItem(Constants.sessionStorageErrorRedirectTimeKey)
-                window.location.replace(Utils.getFrontendBaseURL(true))
-                return
-            }
         }
 
         const url = Utils.getFrontendBaseURL(true) + '/error?id=unknown'
--- a/webapp/src/error_boundary.tsx
+++ b/webapp/src/error_boundary.tsx
@@ -19,7 +19,9 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     msg = 'Redirecting to error page...'
 
     handleError = (): void => {
-        if (window.location.pathname.endsWith('/error')) {
+        const frontendBase = Utils.getFrontendBaseURL()
+        const expectedErrorPathname = frontendBase ? `/${frontendBase}/error` : '/error'
+        if (window.location.pathname === expectedErrorPathname) {
             return
         }
 
--- a/webapp/src/error_boundary.tsx
+++ b/webapp/src/error_boundary.tsx
@@ -19,12 +19,6 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     msg = 'Redirecting to error page...'
 
     handleError = (): void => {
-        const frontendBase = Utils.getFrontendBaseURL()
-        const expectedErrorPathname = frontendBase ? `/${frontendBase}/error` : '/error'
-        if (window.location.pathname === expectedErrorPathname) {
-            return
-        }
-
         const url = Utils.getFrontendBaseURL(true) + '/error?id=unknown'
         window.location.replace(url)
     }
--- a/webapp/src/pages/boardPage/boardPage.tsx
+++ b/webapp/src/pages/boardPage/boardPage.tsx
@@ -65,7 +65,7 @@ import WebsocketConnection from './websocketConnection'
 
 import './boardPage.scss'
 
-const MALFORMED_URL_PATTERNS = ['/error', 'plugins/', 'boards/', 'team/']
+const MALFORMED_URL_SEGMENTS = new Set(['error', 'plugins', 'boards', 'team'])
 
 type Props = {
     readonly?: boolean
@@ -92,18 +92,21 @@ const BoardPage = (props: Props): JSX.Element => {
 
     // Synchronous URL validation - runs before any data loading effects
     const isUrlMalformed = useMemo(() => {
-        const boardId = match.params.boardId
-        if (boardId && MALFORMED_URL_PATTERNS.some(pattern => boardId.includes(pattern))) {
+        const {boardId, viewId: viewIdParam, cardId} = match.params
+        if (boardId && MALFORMED_URL_SEGMENTS.has(boardId)) {
             Utils.logWarn(`Detected malformed boardId in URL: ${boardId}`)
             return true
         }
-        const viewIdParam = match.params.viewId
-        if (viewIdParam && MALFORMED_URL_PATTERNS.some(pattern => viewIdParam.includes(pattern))) {
+        if (viewIdParam && MALFORMED_URL_SEGMENTS.has(viewIdParam)) {
             Utils.logWarn(`Detected malformed viewId in URL: ${viewIdParam}`)
             return true
         }
+        if (cardId && MALFORMED_URL_SEGMENTS.has(cardId)) {
+            Utils.logWarn(`Detected malformed cardId in URL: ${cardId}`)
+            return true
+        }
         return false
-    }, [match.params.boardId, match.params.viewId])
+    }, [match.params.boardId, match.params.viewId, match.params.cardId])
 
     // if we're in a legacy route and not showing a shared board,
     // redirect to the new URL schema equivalent
--- a/webapp/src/pages/boardPage/boardPage.tsx
+++ b/webapp/src/pages/boardPage/boardPage.tsx
@@ -4,7 +4,7 @@
 import React, {useEffect, useState, useMemo, useCallback} from 'react'
 import {batch} from 'react-redux'
 import {FormattedMessage, useIntl} from 'react-intl'
-import {useRouteMatch, useHistory} from 'react-router-dom'
+import {Redirect, useRouteMatch, useHistory} from 'react-router-dom'
 
 import Workspace from '../../components/workspace'
 import VersionMessage from '../../components/messages/versionMessage'
@@ -65,6 +65,8 @@ import WebsocketConnection from './websocketConnection'
 
 import './boardPage.scss'
 
+const MALFORMED_URL_PATTERNS = ['/error', 'plugins/', 'boards/', 'team/']
+
 type Props = {
     readonly?: boolean
     new?: boolean
@@ -88,36 +90,20 @@ const BoardPage = (props: Props): JSX.Element => {
     const history = useHistory()
     const globalError = useAppSelector<string>(getGlobalError)
 
-    // Early parameter validation to prevent errors from malformed URLs
-    // This runs before any data loading or rendering logic
-    useEffect(() => {
-        // Validate boardId if present - should be a valid ID format (not contain 'error', 'plugins', etc.)
+    // Synchronous URL validation - runs before any data loading effects
+    const isUrlMalformed = useMemo(() => {
         const boardId = match.params.boardId
-        if (boardId) {
-            // Check if boardId looks malformed (contains path segments like 'error', 'plugins', 'boards', etc.)
-            const malformedPatterns = ['/error', 'plugins/', 'boards/', 'team/']
-            const isMalformed = malformedPatterns.some(pattern => boardId.includes(pattern))
-
-            if (isMalformed) {
-                Utils.logWarn(`Detected malformed boardId in URL: ${boardId}`)
-                history.replace('/error?id=unknown')
-                return
-            }
+        if (boardId && MALFORMED_URL_PATTERNS.some(pattern => boardId.includes(pattern))) {
+            Utils.logWarn(`Detected malformed boardId in URL: ${boardId}`)
+            return true
         }
-
-        // Validate viewId if present
         const viewIdParam = match.params.viewId
-        if (viewIdParam) {
-            const malformedPatterns = ['/error', 'plugins/', 'boards/', 'team/']
-            const isMalformed = malformedPatterns.some(pattern => viewIdParam.includes(pattern))
-
-            if (isMalformed) {
-                Utils.logWarn(`Detected malformed viewId in URL: ${viewIdParam}`)
-                history.replace('/error?id=unknown')
-                return
-            }
+        if (viewIdParam && MALFORMED_URL_PATTERNS.some(pattern => viewIdParam.includes(pattern))) {
+            Utils.logWarn(`Detected malformed viewId in URL: ${viewIdParam}`)
+            return true
         }
-    }, [match.params.boardId, match.params.viewId, history])
+        return false
+    }, [match.params.boardId, match.params.viewId])
 
     // if we're in a legacy route and not showing a shared board,
     // redirect to the new URL schema equivalent
@@ -193,6 +179,7 @@ const BoardPage = (props: Props): JSX.Element => {
         }
 
         const dispatchLoadAction = () => {
+            if (isUrlMalformed || !match.params.boardId) return
             dispatch(loadAction(match.params.boardId))
         }
 
@@ -220,7 +207,7 @@ const BoardPage = (props: Props): JSX.Element => {
             wsClient.removeOnChange(incrementalBoardMemberUpdate, 'boardMembers')
             wsClient.removeOnReconnect(dispatchLoadAction)
         }
-    }, [me?.id, activeBoardId])
+    }, [me?.id, activeBoardId, isUrlMalformed, match.params.boardId])
 
     const onConfirmJoin = async () => {
         if (me) {
@@ -301,6 +288,7 @@ const BoardPage = (props: Props): JSX.Element => {
     }, [])
 
     useEffect(() => {
+        if (isUrlMalformed) return
         dispatch(loadAction(match.params.boardId))
 
         if (match.params.boardId) {
@@ -316,13 +304,12 @@ const BoardPage = (props: Props): JSX.Element => {
                 }
             }
         }
-    }, [teamId, match.params.boardId, viewId, me?.id])
+    }, [teamId, match.params.boardId, viewId, me?.id, isUrlMalformed])
 
     useEffect(() => {
-        if (match.params.boardId && !props.readonly && me) {
-            loadOrJoinBoard(me, teamId, match.params.boardId)
-        }
-    }, [teamId, match.params.boardId, me?.id])
+        if (isUrlMalformed || !match.params.boardId || props.readonly || !me) return
+        loadOrJoinBoard(me, teamId, match.params.boardId)
+    }, [teamId, match.params.boardId, me?.id, isUrlMalformed])
 
     // Track when board has been loaded at least once
     useEffect(() => {
@@ -334,10 +321,9 @@ const BoardPage = (props: Props): JSX.Element => {
     // When the board is removed from the store while viewing (e.g. user was removed via websocket),
     // re-verify access so we show access-denied instead of the template picker
     useEffect(() => {
-        if (match.params.boardId && !props.readonly && me && !currentBoard && boardWasLoaded) {
-            loadOrJoinBoard(me, teamId, match.params.boardId)
-        }
-    }, [teamId, match.params.boardId, me?.id, currentBoard, loadOrJoinBoard, boardWasLoaded])
+        if (isUrlMalformed || !match.params.boardId || props.readonly || !me || currentBoard || !boardWasLoaded) return
+        loadOrJoinBoard(me, teamId, match.params.boardId)
+    }, [teamId, match.params.boardId, me?.id, currentBoard, loadOrJoinBoard, boardWasLoaded, isUrlMalformed])
 
     const handleUnhideBoard = async (boardID: string) => {
         if (!me || !category) {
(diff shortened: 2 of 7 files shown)


pvev:
@coderabbitai review

Rajat-Dabade:
@coderabbitai review this PR.

Rajat-Dabade:
@coderabbitai regarding this comment: 
> Session flag not cleared, may persist across navigation.
The sessionStorageIgnoreStoredUrlsKey flag is read here but not removed. While teamToBoardAndViewRedirect.tsx clears it, if a user navigates to / and this redirect fires, they land on /team/${teamID} without ever hitting TeamToBoardAndViewRedirect. Subsequent navigations to / would still skip the stored URL logic until the flag is cleared elsewhere.
Consider clearing the flag after reading it here to ensure it only takes effect once.

Removing the flag before the <Redirect> causes a re-render loop. React can re-render the route component before the redirect completes, and on the second render, the flag is gone, so lastBoardId (pointing to the broken board) is used again, sending the user back to the error page. sessionStorage is already tab-scoped and cleared on tab close or fresh navigation, so the flag naturally expires after the session. No change needed here.

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -65,6 +65,8 @@ import WebsocketConnection from './websocketConnection'
 
 import './boardPage.scss'
 
+const MALFORMED_URL_PATTERNS = ['/error', 'plugins/', 'boards/', 'team/']
```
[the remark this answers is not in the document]
Rajat-Dabade: Done.

devinbinnie (commented):
I'm a bit confused about the approach here. It seems like the crux of the issue is that we're trying to redirect to `plugins/focalboard` as a relative URL to begin with, which it doesn't seem like we're addressing. I'd rather not trying to catch it by trying to count the number of redirects or detecting specific URL pieces.

If there is a root cause of this issue and it is simply that we're not redirecting correctly, I'd like to address it there. Also, I can't seem to reproduce the original path in the ticket anymore myself, is there a way we can reproduce this infinite loop still?

Rajat-Dabade:
@devinbinnie, sorry for the confusion. 

Here are the steps to reproduce this issue: 

- Create a board.
- Select any template.
- Add a multiperson property to the card and assign some member to it in 3 cards. 
- Now sort the board with multiperson property. 
- Hard reload on the board. 
- See the error. It will loop. 

The above issue is fixed using this PR: https://github.com/mattermost/mattermost-plugin-boards/pull/176

This was because https://hub.mattermost.com/pde/pl/dt7zoqe5ptngt8f7sng1gwtc3w 

Now, the concern was that if, in the future, the same thing happens for another reason and the error boundary is hit, the board should not redirect to itself; therefore, we implemented a defensive mechanism in this PR. 

But the defensive mechanisms address a secondary loop that happens after that initial fix, which was
- User encounters a bug in a board (e.g., the multiperson sorting issue that is now fixed)
- Error boundary catches it → redirects to /boards/error (absolute URL works)
- User is on the error page and clicks the "Back to Home" button
- Router's HomeToCurrentTeam automatically redirects to the last visited board (stored in `UserSettings.lastBoardId`)
- That board still has the same bug -> error boundary triggers again
- Back to step 2 -> infinite redirect loop (not URL growth, but a redirect cycle)

Why the Defensive Mechanisms:
The loop detection and ignore flags prevent this cycle by:
- Clearing local stored board/view IDs when returning from the error page
- Setting a flag to skip auto-redirect to the broken board
- Detecting if we're stuck in a redirect loop (>3 redirects in 10 seconds)

The absolute URL fix prevents the first issue (URL path growth), but doesn't prevent the router from sending it back to a broken board. That is what the defensive mechanisms handle.

Let me know your thoughts on this one. Thank you.

devinbinnie:
> @devinbinnie, sorry for the confusion.
> 
> Here are the steps to reproduce this issue:
> 
>     * Create a board.
> 
>     * Select any template.
> 
>     * Add a multiperson property to the card and assign some member to it in 3 cards.
> 
>     * Now sort the board with multiperson property.
> 
>     * Hard reload on the board.
> 
>     * See the error. It will loop.
> 
> 
> The above issue is fixed using this PR: #176
> 
> This was because https://hub.mattermost.com/pde/pl/dt7zoqe5ptngt8f7sng1gwtc3w
> 
> Now, the concern was that if, in the future, the same thing happens for another reason and the error boundary is hit, the board should not redirect to itself; therefore, we implemented a defensive mechanism in this PR.
> 
> But the defensive mechanisms address a secondary loop that happens after that initial fix, which was
> 
>     * User encounters a bug in a board (e.g., the multiperson sorting issue that is now fixed)
> 
>     * Error boundary catches it → redirects to /boards/error (absolute URL works)
> 
>     * User is on the error page and clicks the "Back to Home" button
> 
>     * Router's HomeToCurrentTeam automatically redirects to the last visited board (stored in `UserSettings.lastBoardId`)
> 
>     * That board still has the same bug -> error boundary triggers again
> 
>     * Back to step 2 -> infinite redirect loop (not URL growth, but a redirect cycle)
> 
> 
> Why the Defensive Mechanisms: The loop detection and ignore flags prevent this cycle by:
> 
>     * Clearing local stored board/view IDs when returning from the error page
> 
>     * Setting a flag to skip auto-redirect to the broken board
> 
>     * Detecting if we're stuck in a redirect loop (>3 redirects in 10 seconds)
> 
> 
> The absolute URL fix prevents the first issue (URL path growth), but doesn't prevent the router from sending it back to a broken board. That is what the defensive mechanisms handle.
> 
> Let me know your thoughts on this one. Thank you.

Thanks for the context. 

Let me know if I'm misunderstanding this, but I'm a bit confused why we decide to redirect to an error page on any JavaScript error. This seems like no matter what if there's a bug in the main board, it will cause this loop. Is it any JavaScript error that will trigger this? Do we know the reason why we went with that design to begin with?

The way I would see if is if we were going to have an error boundary for all JavaScript, the intention of that is to make sure the UI we fall back to is safe. The idea of using the same method as the web app where we just redirect back to the board seems unsafe since that's likely where the error was caused.

I think the way I would solve this would be if a JavaScript error triggered this flow, to redirect to the safest possible place, which I would guess is the "Create a board" screen? Another option would be to just let it white screen like the web app does, but I guess the error page is more elegant.

I feel that trying to catch this redirect loop just puts a band-aid on top of what might be a bigger bug, since users would just hit go back to team enough times until it works and might not report the issue. If there was a way to detect immediately that the board caused the problem without forcing just an arbitrary count, then I'd be willing to explore that.

But I think the current approach might do more harm than good, so I'm thinking if we want to remedy this, we might need to re-examine the need for that error boundary. One thing I'd like to think about is what might happen for smaller JS errors that aren't as consequential?

Rajat-Dabade:
> > @devinbinnie, sorry for the confusion.
> > Here are the steps to reproduce this issue:
> > ```
> > * Create a board.
> > 
> > * Select any template.
> > 
> > * Add a multiperson property to the card and assign some member to it in 3 cards.
> > 
> > * Now sort the board with multiperson property.
> > 
> > * Hard reload on the board.
> > 
> > * See the error. It will loop.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The above issue is fixed using this PR: #176
> > This was because [hub.mattermost.com/pde/pl/dt7zoqe5ptngt8f7sng1gwtc3w](https://hub.mattermost.com/pde/pl/dt7zoqe5ptngt8f7sng1gwtc3w)
> > Now, the concern was that if, in the future, the same thing happens for another reason and the error boundary is hit, the board should not redirect to itself; therefore, we implemented a defensive mechanism in this PR.
> > But the defensive mechanisms address a secondary loop that happens after that initial fix, which was
> > ```
> > * User encounters a bug in a board (e.g., the multiperson sorting issue that is now fixed)
> > 
> > * Error boundary catches it → redirects to /boards/error (absolute URL works)
> > 
> > * User is on the error page and clicks the "Back to Home" button
> > 
> > * Router's HomeToCurrentTeam automatically redirects to the last visited board (stored in ``UserSettings.lastBoardId``)
> > 
> > * That board still has the same bug -> error boundary triggers again
> > 
> > * Back to step 2 -> infinite redirect loop (not URL growth, but a redirect cycle)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Why the Defensive Mechanisms: The loop detection and ignore flags prevent this cycle by:
> > ```
> > * Clearing local stored board/view IDs when returning from the error page
> > 
> > * Setting a flag to skip auto-redirect to the broken board
> > 
> > * Detecting if we're stuck in a redirect loop (>3 redirects in 10 seconds)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The absolute URL fix prevents the first issue (URL path growth), but doesn't prevent the router from sending it back to a broken board. That is what the defensive mechanisms handle.
> > Let me know your thoughts on this one. Thank you.
> 
> Thanks for the context.
> 
> Let me know if I'm misunderstanding this, but I'm a bit confused why we decide to redirect to an error page on any JavaScript error. This seems like no matter what if there's a bug in the main board, it will cause this loop. Is it any JavaScript error that will trigger this? Do we know the reason why we went with that design to begin with?
> 
> The way I would see if is if we were going to have an error boundary for all JavaScript, the intention of that is to make sure the UI we fall back to is safe. The idea of using the same method as the web app where we just redirect back to the board seems unsafe since that's likely where the error was caused.
> 
> I think the way I would solve this would be if a JavaScript error triggered this flow, to redirect to the safest possible place, which I would guess is the "Create a board" screen? Another option would be to just let it white screen like the web app does, but I guess the error page is more elegant.
> 
> I feel that trying to catch this redirect loop just puts a band-aid on top of what might be a bigger bug, since users would just hit go back to team enough times until it works and might not report the issue. If there was a way to detect immediately that the board caused the problem without forcing just an arbitrary count, then I'd be willing to explore that.
> 
> But I think the current approach might do more harm than good, so I'm thinking if we want to remedy this, we might need to re-examine the need for that error boundary. One thing I'd like to think about is what might happen for smaller JS errors that aren't as consequential?

@devinbinnie You are right that redirecting back to the error page -> "Back to Home" -> auto-redirect to broken board creates a problematic loop. The current count-redirects approach is indeed a band-aid. Also, the error boundary was already in the codebase for a very long, it catches any unhandled React error and redirects to `/error?id=unknown`. This is a common React pattern to prevent the entire app from white-screening.

Now that you said, the better alternative that I can think of is: 

Option 1: Clear the broken board from history immediately
```javascript
   // When an error occurs, clear the lastBoardId
  localStorage.removeItem(UserSettingKey.LastBoardId)
  localStorage.removeItem(UserSettingKey.LastViewId)
  // Then redirect to the team page
  window.location.replace(Utils.getFrontendBaseURL(true) + `/team/${teamId}`)
```

  Option 2: Store the failing board ID and skip it
```javascript  
  // Store which board caused the error
  sessionStorage.setItem('board_error_board_id', currentBoardId)
  // In TeamToBoardAndViewRedirect, check and skip that board
  if (boardID === sessionStorage.getItem('board_error_board_id')) {
    // Skip to next available board or team page
  }
```

We can go with option 2 
  - Users often work across multiple boards
  - If Board A errors, but Boards B, C, and D are fine, Option 3 lets them continue
  - Option 1 forces them to start from scratch every time

Also, from a future perspective, we can track which boards are causing errors and might help in finding the root cause as well. 

Thoughts?

devinbinnie:
> > > @devinbinnie, sorry for the confusion.
> > > Here are the steps to reproduce this issue:
> > > ```
> > > * Create a board.
> > > 
> > > * Select any template.
> > > 
> > > * Add a multiperson property to the card and assign some member to it in 3 cards.
> > > 
> > > * Now sort the board with multiperson property.
> > > 
> > > * Hard reload on the board.
> > > 
> > > * See the error. It will loop.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The above issue is fixed using this PR: #176
> > > This was because [hub.mattermost.com/pde/pl/dt7zoqe5ptngt8f7sng1gwtc3w](https://hub.mattermost.com/pde/pl/dt7zoqe5ptngt8f7sng1gwtc3w)
> > > Now, the concern was that if, in the future, the same thing happens for another reason and the error boundary is hit, the board should not redirect to itself; therefore, we implemented a defensive mechanism in this PR.
> > > But the defensive mechanisms address a secondary loop that happens after that initial fix, which was
> > > ```
> > > * User encounters a bug in a board (e.g., the multiperson sorting issue that is now fixed)
> > > 
> > > * Error boundary catches it → redirects to /boards/error (absolute URL works)
> > > 
> > > * User is on the error page and clicks the "Back to Home" button
> > > 
> > > * Router's HomeToCurrentTeam automatically redirects to the last visited board (stored in ``UserSettings.lastBoardId``)
> > > 
> > > * That board still has the same bug -> error boundary triggers again
> > > 
> > > * Back to step 2 -> infinite redirect loop (not URL growth, but a redirect cycle)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Why the Defensive Mechanisms: The loop detection and ignore flags prevent this cycle by:
> > > ```
> > > * Clearing local stored board/view IDs when returning from the error page
> > > 
> > > * Setting a flag to skip auto-redirect to the broken board
> > > 
> > > * Detecting if we're stuck in a redirect loop (>3 redirects in 10 seconds)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The absolute URL fix prevents the first issue (URL path growth), but doesn't prevent the router from sending it back to a broken board. That is what the defensive mechanisms handle.
> > > Let me know your thoughts on this one. Thank you.
> > 
> > 
> > Thanks for the context.
> > Let me know if I'm misunderstanding this, but I'm a bit confused why we decide to redirect to an error page on any JavaScript error. This seems like no matter what if there's a bug in the main board, it will cause this loop. Is it any JavaScript error that will trigger this? Do we know the reason why we went with that design to begin with?
> > The way I would see if is if we were going to have an error boundary for all JavaScript, the intention of that is to make sure the UI we fall back to is safe. The idea of using the same method as the web app where we just redirect back to the board seems unsafe since that's likely where the error was caused.
> > I think the way I would solve this would be if a JavaScript error triggered this flow, to redirect to the safest possible place, which I would guess is the "Create a board" screen? Another option would be to just let it white screen like the web app does, but I guess the error page is more elegant.
> > I feel that trying to catch this redirect loop just puts a band-aid on top of what might be a bigger bug, since users would just hit go back to team enough times until it works and might not report the issue. If there was a way to detect immediately that the board caused the problem without forcing just an arbitrary count, then I'd be willing to explore that.
> > But I think the current approach might do more harm than good, so I'm thinking if we want to remedy this, we might need to re-examine the need for that error boundary. One thing I'd like to think about is what might happen for smaller JS errors that aren't as consequential?
> 
> @devinbinnie You are right that redirecting back to the error page -> "Back to Home" -> auto-redirect to broken board creates a problematic loop. The current count-redirects approach is indeed a band-aid. Also, the error boundary was already in the codebase for a very long, it catches any unhandled React error and redirects to `/error?id=unknown`. This is a common React pattern to prevent the entire app from white-screening.
> 
> Now that you said, the better alternative that I can think of is:
> 
> Option 1: Clear the broken board from history immediately
> 
> ```js
>    // When an error occurs, clear the lastBoardId
>   localStorage.removeItem(UserSettingKey.LastBoardId)
>   localStorage.removeItem(UserSettingKey.LastViewId)
>   // Then redirect to the team page
>   window.location.replace(Utils.getFrontendBaseURL(true) + `/team/${teamId}`)
> ```
> 
> Option 2: Store the failing board ID and skip it
> 
> ```js
>   // Store which board caused the error
>   sessionStorage.setItem('board_error_board_id', currentBoardId)
>   // In TeamToBoardAndViewRedirect, check and skip that board
>   if (boardID === sessionStorage.getItem('board_error_board_id')) {
>     // Skip to next available board or team page
>   }
> ```
> 
> We can go with option 2
> 
>     * Users often work across multiple boards
> 
>     * If Board A errors, but Boards B, C, and D are fine, Option 3 lets them continue
> 
>     * Option 1 forces them to start from scratch every time
> 
> 
> Also, from a future perspective, we can track which boards are causing errors and might help in finding the root cause as well.
> 
> Thoughts?

Option 1 actually sounds like the better option here, since it would cover all cases and more than likely return the user to a safe place.

We could store the failing board sure - but once it's fixed how does the user clear it? It's not a good expectation to get them to clear `sessionStorage` since most users don't know what that is.

As a follow-up question: why would one Board be working and another be broken in general? Is it just because a specific bug is affecting one Board due to the way it's set up? Or is there a problem with the data underneath? I'm trying to understand exactly what we're trying to avoid here. If it's just to avoid bugs - I don't think we'd want to outright stop users from seeing their data.

webapp/src/error_boundary.tsx
```diff
@@ -19,8 +20,33 @@ export default class ErrorBoundary extends React.Component<Props, State> {
...
+            sessionStorage.setItem(Constants.sessionStorageErrorRedirectTimeKey, now.toString())
+
+            // If redirected more than 3 times in 10 seconds, go to boards root
+            if (newCount > 3) {
+                sessionStorage.removeItem(Constants.sessionStorageErrorRedirectCountKey)
+                sessionStorage.removeItem(Constants.sessionStorageErrorRedirectTimeKey)
+                window.location.replace(Utils.getFrontendBaseURL(true))
+                return
+            }
+        }
```
marianunez: Is this necessary? We should provide a fix to not redirect so this should tracking with a max allowed seems hacky

webapp/src/error_boundary.tsx
```diff
@@ -19,8 +20,33 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     msg = 'Redirecting to error page...'
 
     handleError = (): void => {
-        const url = Utils.getBaseURL() + '/error?id=unknown'
```
marianunez, line 22: Why is this returning no full path base url? Seems that we need to understand this behavior of why it's working with a relative path and why it's not handled properly

webapp/src/error_boundary.tsx
marianunez: Do we need this looping detection if we fix it properly? We shouldn't need limits or tracking here if it works as expected - this seems a bit hacky as this doesn't seem to be a case of a race condition

webapp/src/error_boundary.tsx
```diff
@@ -19,8 +20,33 @@ export default class ErrorBoundary extends React.Component<Props, State> {
...
-        Utils.log('error boundary redirecting to ' + url)
+        if (window.location.pathname.endsWith('/error')) {
```
marianunez: why would it be handling an error in the error page?

marianunez (changes requested):

webapp/src/error_boundary.tsx
Rajat-Dabade: Agreed, we are implementing a new way to resolve this issue, so this will be removed.

Rajat-Dabade: No longer required. Will remove it.

webapp/src/error_boundary.tsx
Rajat-Dabade: So the ErrorBoundary component wraps the whole application, including the error page itself. In the normal case, if something breaks on a board page, the ErrorBoundary catches it and redirects the user to `/boards/error`, and this causes a full page reload, everything resets, and the error page loads normally. The user just sees a proper error screen and can continue without issues.

The problem is when the error page itself has a bug. Without this safety check, the ErrorBoundary will keep redirecting to `/boards/error` even though we are already there. This ends up creating an infinite loop where the page keeps reloading again and again. The pathname check avoids this by not redirecting if we’re already on the error page.

webapp/src/error_boundary.tsx
Rajat-Dabade: The relative path doesn't work properly. This is what happens with a relative path:
`Utils.getBaseURL()` returns "plugins/focalboard" (no leading /). When the error boundary redirects from a board page:

Current URL: /boards/team/abc/board123/view456
window.location.replace('plugins/focalboard/error?id=unknown')
No leading / = browser treats as relative to current path
Appends to the current URL instead of navigating to the root:
Result: `/boards/team/abc/board123/view456/plugins/focalboard/error`

This matches the wrong route, throws another error, and creates an infinite loop with the URL growing each time:
  .../plugins/focalboard/error
  .../plugins/focalboard/plugins/focalboard/error
  .../plugins/focalboard/plugins/focalboard/plugins/focalboard/error
  ...

The fix: Using an absolute URL ensures we navigate to the exact error page route instead of appending to the current path:
window.location.replace('http://localhost:8065/boards/error?id=unknown')
Result: /boards/error

Rajat-Dabade:
> As a follow-up question: why would one Board be working and another be broken in general? Is it just because a specific bug is affecting one Board due to the way it's set up? Or is there a problem with the data underneath? I'm trying to understand exactly what we're trying to avoid here. If it's just to avoid bugs - I don't think we'd want to outright stop users from seeing their data.

@devinbinnie  Good question! This change is not hiding any data from users in fact, it helps them access more of it. The problem was that some boards can have data-specific bugs, and if a user’s “last visited board” is one of those, they get stuck in loop. Clicking “Back to Home” on the error page would just keep redirecting them back to the same broken board, blocking access to all their other boards.

This fix gives users a way out. Now, when they click “Back to Home,” they actually reach the home page and can access all their other boards normally. The skipRedirect flag introduced in the new changes (Option 1 implementation) is just a escape, it stops the automatic redirect once, and resets when the user manually picks a board. Nothing is hidden, it just prevents users from getting stuck.

devinbinnie:
> > As a follow-up question: why would one Board be working and another be broken in general? Is it just because a specific bug is affecting one Board due to the way it's set up? Or is there a problem with the data underneath? I'm trying to understand exactly what we're trying to avoid here. If it's just to avoid bugs - I don't think we'd want to outright stop users from seeing their data.
> 
> @devinbinnie Good question! This change is not hiding any data from users in fact, it helps them access more of it. The problem was that some boards can have data-specific bugs, and if a user’s “last visited board” is one of those, they get stuck in loop. Clicking “Back to Home” on the error page would just keep redirecting them back to the same broken board, blocking access to all their other boards.
> 
> This fix gives users a way out. Now, when they click “Back to Home,” they actually reach the home page and can access all their other boards normally. The skipRedirect flag introduced in the new changes (Option 1 implementation) is just a escape, it stops the automatic redirect once, and resets when the user manually picks a board. Nothing is hidden, it just prevents users from getting stuck.

So is this a problem with the data itself (ie. the data in the DB is malformed somehow) or a problem with the way Boards is handing the data (ie. a bug in the plugin code)?

Rajat-Dabade:
> So is this a problem with the data itself (ie. the data in the DB is malformed somehow) or a problem with the way Boards is handing the data (ie. a bug in the plugin code)?

@devinbinnie This isn't an issue with bad data, it’s actually a bug in the plugin code. For example, the bug we fixed in this PR: https://github.com/mattermost/mattermost-plugin-boards/pull/176 in the `.localeCompare` case, the data is completely valid, but the code assumes everything is a string and crashes when it encounters arrays or objects.

That said, this fix is not tied to any one specific bug. The real problem is that if a user’s “last visited board” triggers any kind of frontend error, they get stuck in a redirect loop and can’t access anything else. The `skipRedirect` solution acts like a safety escape. When users click “Back to Home” from the error page, we skip the auto-redirect and let them reach the home page. From there, they can open other boards that work fine. It’s basically a safety net that prevents users from getting stuck, no matter what caused the crash.

marianunez:
@coderabbitai do a full re-review

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -88,6 +90,24 @@ const BoardPage = (props: Props): JSX.Element => {
...
+        if (viewIdParam && MALFORMED_URL_SEGMENTS.has(viewIdParam)) {
+            Utils.logWarn(`Detected malformed viewId in URL: ${viewIdParam}`)
+            return true
+        }
+        if (cardId && MALFORMED_URL_SEGMENTS.has(cardId)) {
+            Utils.logWarn(`Detected malformed cardId in URL: ${cardId}`)
+            return true
+        }
+        return false
+    }, [match.params.boardId, match.params.viewId, match.params.cardId])
```
marianunez: I understand this is defensive check but this should never happen if we fix the underlying routing issues. 

1/5 that we don't include this as it may mask other routing issues and could introduce regressions (an example is the case CodeRabbit is highlighting below).

webapp/src/error_boundary.tsx
marianunez: > The problem is when the error page itself has a bug.

What type of bug would that be? The error page should be a simple page so this seems also that we are patching in case the routing bug is not fixed.

1/5 we don't include this and let any other errors surface

marianunez (approved):
Approving as this solves the issue but 1/5 we should remove all extra "hardening" that we included in case routing fails again. We should be more confident on the root problem fix that they should not be needed.

webapp/src/error_boundary.tsx
```diff
@@ -19,8 +19,11 @@ export default class ErrorBoundary extends React.Component<Props, State> {
...
+            return
+        }
+
+        const url = Utils.getFrontendBaseURL(true) + '/error?id=unknown'
         window.location.replace(url)
```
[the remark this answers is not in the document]
Rajat-Dabade: Agreed.

webapp/src/pages/boardPage/boardPage.tsx
Rajat-Dabade: Agree, this is defensive and maybe can lead to routing bugs. Happy to remove it and rely on fixing routing at the source. I will verify that error handling still behaves the same, and then will remove it.

webapp/src/error_boundary.tsx
Rajat-Dabade: Agree, I mean, it's just a defensive code, for any bugs in the error_boundary.tsx. But we can remove it.

Rajat-Dabade:
@coderabbitai can you take a final pass, and if everything looks good, approve it.

Rajat-Dabade:
@andrleite, I see you have a commit on this PR with `id-token: write` right permission. I have already taken care of

```yml
permissions:
  id-token: write
```

in this PR: https://github.com/mattermost/mattermost-plugin-boards/pull/155

We should revert the changes from this PR, because the merge will be blocked because of the failing webapp-test.

Rajat-Dabade:
@coderabbitai Confirmed that `fetchTeams` does not populate `myBoards`, boards come from `initialLoad`  board flows, so `myBoards` can be [] while `teamsFetched` is already true. `!validBoardIds.has(lastBoardID)` is true for an empty set, so we were incorrectly clearing `lastBoardId` on first paint. Fixed by only running invalid-board fallback when `validBoardIds.size > 0 && !validBoardIds.has(lastBoardID)`.

PR #179, opened 2026-03-17
Validate file ownership before associating with a block

Rajat-Dabade:
#### Summary
When a file is uploaded to a board, we now store the board’s ID in the `FileInfo.Path`. This helps us track which board the file belongs to right from the time it’s uploaded.

##### What was fixed?
File storage structure: 
Files are now stored in a way that clearly shows which board they belong to:

Before: 
```
boards/YYYYMMDD/{filename}
```

Now: 
```
boards/YYYYMMDD/{boardID}/{filename}
```
This means:
- Each file is linked to a board through its path
- No need for extra database fields to track ownership

(Template files are unchanged and still follow their own format.)

##### How validation works
Before creating or updating a block, we now check that any attached file actually belongs to the same board.
This validation is applied in all block write operations:
- InsertBlocksAndNotify
- InsertBlockAndNotify
- PatchBlockAndNotify
- PatchBlocksAndNotify (batch)

For the file stored in an older format: 

```
boards/YYYYMMDD/{filename}
```
We already have the function `validateFileReferencedByBoard`, this will
- First, find the block associated with the file.
- Find the boardId associated with the block.
- If a block is found and it belongs to the same board as the incoming request, then allow
- If a block is found, but it belongs to a different board, then reject
- If no block is found yet (fresh upload, not yet attached to any block), then allow. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67760

## Change Impact: 🔴 High

**Reasoning:** The change enforces board-scoped ownership validation across all block write flows and adds a startup data migration that moves file objects and updates DB paths; this affects authorization, file storage layout, and data migration/persistence across multiple modules (files, blocks, store/migration) and touches critical user-facing flows.

**Regression Risk:** High — these changes modify authorization checks for block mutations, introduce a migration that moves and updates FileInfo paths, and adjust storage paths for uploaded files. Gaps in tests (missing FileExists mocks for CopyCardFiles scenarios) and synchronous migration-on-startup behavior increase the risk of regressions or startup delays; fallback/compatibility logic mitigates some transitional cases but does not eliminate the potential for incorrectly rejected requests or missed ownership detection.

**QA Recommendation:** Require thorough manual and integration testing, including:
1) End-to-end tests for file upload, copy/clone flows (CopyCardFiles), block create/patch/insert (single and batch) verifying ownership validation succeeds/fails appropriately.
2) Migration testing on large datasets and cold-start behavior (including scenarios where migration is skipped) to measure runtime and ensure no data loss or inconsistent FileInfo.Path states.
3) Tests for legacy files (3-part paths and empty PostId) and template paths to verify fallback behavior.
4) Add/adjust unit tests to mock FileExists where CopyFile/FileExists are used so test behavior matches runtime code paths.
5) Verify notification and batch-patch ordering assumptions (map-by-ID fixes) and watch for performance hotspots (repeated GetBoard calls, history fetches during filtering).

*Generated by CodeRabbitAI*

### DIFF
--- a/server/app/blocks.go
+++ b/server/app/blocks.go
@@ -82,7 +82,8 @@ func (a *App) PatchBlockAndNotify(blockID string, blockPatch *model.BlockPatch, 
 
 	// Reject patches that reference files not belonging to this board.
 	if blockPatch.UpdatedFields != nil {
-		if err := a.validateFileRefsInFields(board.TeamID, oldBlock.BoardID, blockID, blockPatch.UpdatedFields); err != nil {
+		err = a.validateFileRefsInFields(board.TeamID, oldBlock.BoardID, blockID, blockPatch.UpdatedFields)
+		if err != nil {
 			return nil, err
 		}
 	}
--- a/server/app/blocks.go
+++ b/server/app/blocks.go
@@ -82,15 +82,8 @@ func (a *App) PatchBlockAndNotify(blockID string, blockPatch *model.BlockPatch, 
 
 	// Reject patches that reference files not belonging to this board.
 	if blockPatch.UpdatedFields != nil {
-		if fileID, ok := blockPatch.UpdatedFields[model.BlockFieldFileId].(string); ok && fileID != "" {
-			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, oldBlock.BoardID, fileID); ownershipErr != nil {
-				return nil, fmt.Errorf("unauthorized file reference in patch for block %s: %w", blockID, ownershipErr)
-			}
-		}
-		if attachmentID, ok := blockPatch.UpdatedFields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
-			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, oldBlock.BoardID, attachmentID); ownershipErr != nil {
-				return nil, fmt.Errorf("unauthorized file reference in patch for block %s: %w", blockID, ownershipErr)
-			}
+		if err := a.validateFileRefsInFields(board.TeamID, oldBlock.BoardID, blockID, blockPatch.UpdatedFields); err != nil {
+			return nil, err
 		}
 	}
 
@@ -161,15 +154,8 @@ func (a *App) PatchBlocksAndNotify(teamID string, blockPatches *model.BlockPatch
 			}
 			boardCache[block.BoardID] = board
 		}
-		if fileID, ok := patch.UpdatedFields[model.BlockFieldFileId].(string); ok && fileID != "" {
-			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, fileID); ownershipErr != nil {
-				return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
-			}
-		}
-		if attachmentID, ok := patch.UpdatedFields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
-			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, attachmentID); ownershipErr != nil {
-				return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
-			}
+		if err := a.validateFileRefsInFields(board.TeamID, block.BoardID, block.ID, patch.UpdatedFields); err != nil {
+			return err
 		}
 	}
 
@@ -195,6 +181,23 @@ func (a *App) PatchBlocksAndNotify(teamID string, blockPatches *model.BlockPatch
 	return nil
 }
 
+// validateFileRefsInFields checks that any file referenced in a block's field
+// map belongs to the given board. It checks both BlockFieldFileId and
+// BlockFieldAttachmentId. blockID is used only for error messages.
+func (a *App) validateFileRefsInFields(teamID, boardID, blockID string, fields map[string]interface{}) error {
+	if fileID, ok := fields[model.BlockFieldFileId].(string); ok && fileID != "" {
+		if err := a.validateFileOwnershipForBlockWrite(teamID, boardID, fileID); err != nil {
+			return fmt.Errorf("unauthorized file reference in block %s: %w", blockID, err)
+		}
+	}
+	if attachmentID, ok := fields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
+		if err := a.validateFileOwnershipForBlockWrite(teamID, boardID, attachmentID); err != nil {
+			return fmt.Errorf("unauthorized file reference in block %s: %w", blockID, err)
+		}
+	}
+	return nil
+}
+
 func (a *App) InsertBlock(block *model.Block, modifiedByID string) error {
 	return a.InsertBlockAndNotify(block, modifiedByID, false)
 }
@@ -205,15 +208,8 @@ func (a *App) InsertBlockAndNotify(block *model.Block, modifiedByID string, disa
 		return bErr
 	}
 
-	if fileID, ok := block.Fields[model.BlockFieldFileId].(string); ok && fileID != "" {
-		if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, fileID); ownershipErr != nil {
-			return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
-		}
-	}
-	if attachmentID, ok := block.Fields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
-		if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, attachmentID); ownershipErr != nil {
-			return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
-		}
+	if err := a.validateFileRefsInFields(board.TeamID, block.BoardID, block.ID, block.Fields); err != nil {
+		return err
 	}
 
 	err := a.store.InsertBlock(block, modifiedByID)
@@ -256,15 +252,8 @@ func (a *App) InsertBlocksAndNotify(blocks []*model.Block, modifiedByID string, 
 
 	// Reject blocks that reference files not belonging to this board.
 	for _, block := range blocks {
-		if fileID, ok := block.Fields[model.BlockFieldFileId].(string); ok && fileID != "" {
-			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, boardID, fileID); ownershipErr != nil {
-				return nil, fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
-			}
-		}
-		if attachmentID, ok := block.Fields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
-			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, boardID, attachmentID); ownershipErr != nil {
-				return nil, fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
-			}
+		if err := a.validateFileRefsInFields(board.TeamID, boardID, block.ID, block.Fields); err != nil {
+			return nil, err
 		}
 	}
 
--- a/server/app/blocks.go
+++ b/server/app/blocks.go
@@ -80,6 +80,20 @@ func (a *App) PatchBlockAndNotify(blockID string, blockPatch *model.BlockPatch, 
 		return nil, err
 	}
 
+	// Reject patches that reference files not belonging to this board.
+	if blockPatch.UpdatedFields != nil {
+		if fileID, ok := blockPatch.UpdatedFields[model.BlockFieldFileId].(string); ok && fileID != "" {
+			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, oldBlock.BoardID, fileID); ownershipErr != nil {
+				return nil, fmt.Errorf("unauthorized file reference in patch for block %s: %w", blockID, ownershipErr)
+			}
+		}
+		if attachmentID, ok := blockPatch.UpdatedFields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
+			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, oldBlock.BoardID, attachmentID); ownershipErr != nil {
+				return nil, fmt.Errorf("unauthorized file reference in patch for block %s: %w", blockID, ownershipErr)
+			}
+		}
+	}
+
 	err = a.store.PatchBlock(blockID, blockPatch, modifiedByID)
 	if err != nil {
 		return nil, err
@@ -122,6 +136,37 @@ func (a *App) PatchBlocksAndNotify(teamID string, blockPatches *model.BlockPatch
 		return err
 	}
 
+	// Build a lookup so we can pair each patch with its block regardless of DB return order.
+	blockByID := make(map[string]*model.Block, len(oldBlocks))
+	for _, b := range oldBlocks {
+		blockByID[b.ID] = b
+	}
+
+	// Reject patches that reference files not belonging to the target board.
+	for i, patch := range blockPatches.BlockPatches {
+		if patch.UpdatedFields == nil {
+			continue
+		}
+		block, ok := blockByID[blockPatches.BlockIDs[i]]
+		if !ok {
+			continue
+		}
+		board, bErr := a.store.GetBoard(block.BoardID)
+		if bErr != nil {
+			return bErr
+		}
+		if fileID, ok := patch.UpdatedFields[model.BlockFieldFileId].(string); ok && fileID != "" {
+			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, fileID); ownershipErr != nil {
+				return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
+			}
+		}
+		if attachmentID, ok := patch.UpdatedFields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
+			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, attachmentID); ownershipErr != nil {
+				return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
+			}
+		}
+	}
+
 	if err := a.store.PatchBlocks(blockPatches, modifiedByID); err != nil {
 		return err
 	}
@@ -152,6 +197,17 @@ func (a *App) InsertBlockAndNotify(block *model.Block, modifiedByID string, disa
 	board, bErr := a.store.GetBoard(block.BoardID)
 	if bErr != nil {
 		return bErr
+	}
+
+	if fileID, ok := block.Fields[model.BlockFieldFileId].(string); ok && fileID != "" {
+		if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, fileID); ownershipErr != nil {
+			return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
+		}
+	}
+	if attachmentID, ok := block.Fields[model.BlockFieldAttachmentId].(string); ok && attachmentID != "" {
+		if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, block.BoardID, attachmentID); ownershipErr != nil {
+			return fmt.Errorf("unauthorized file reference in block %s: %w", block.ID, ownershipErr)
+		}
 	}
 
 	err := a.store.InsertBlock(block, modifiedByID)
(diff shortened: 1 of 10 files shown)


server/app/files.go
```diff
@@ -93,18 +96,56 @@ func (a *App) ValidateFileOwnership(teamID, boardID, filename string) error {
...
 		}
-	} else {
-		if err := a.validateFileReferencedByBoard(boardID, filename); err != nil {
-			return model.NewErrPermission("file does not belong to the specified board")
+	}
+	if err := a.validateFileReferencedByBoard(boardID, filename); err != nil {
+		return model.NewErrPermission("file does not belong to the specified board")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Rajat-Dabade: Done. please re-review

server/app/files.go
```diff
@@ -93,18 +105,58 @@ func (a *App) ValidateFileOwnership(teamID, boardID, filename string) error {
...
+		expectedPath := filepath.ToSlash(filepath.Join(teamID, boardID, filename))
+		if normalizedPath == expectedPath {
+			return nil
+		}
+		return model.NewErrPermission("file does not belong to the specified board")
+	}
+
+	// Regular files: board is stored in PostId at upload time (prefix-stripped to 26 chars).
+	// Empty PostId means the file predates this tracking; allow for backward compatibility.
+	if fileInfo.PostId == "" {
```
devinbinnie: Without the data migration, doesn't this still allows existing files to be read via the same methodology? I think we need to do the data migration in this PR otherwise it would not fix the issue.

server/app/files.go
```diff
@@ -51,6 +60,9 @@ func (a *App) SaveFile(reader io.Reader, teamID, boardID, filename string, asTem
 	fileInfo.Id = getFileInfoID(createdFilename)
 	fileInfo.Path = filePath
 	fileInfo.Size = fileSize
+	// PostId is VARCHAR(26). Boards IDs are 27 chars (1-char type prefix + 26 chars),
+	// so strip the prefix before storing.
+	fileInfo.PostId = stripBoardIDPrefix(boardID)
```
devinbinnie: I do find the idea of using the `PostId` interesting, and looking at it it seems the risks are minimal, but is there any reason we couldn't just go back to including the board ID in the path name? So it would be of the format `boards/YYYYMMDD/{boardID}/{filename}`.

devinbinnie (changes requested):

devinbinnie:
Running it through Claude it also found something we might need to address:

**`CopyCardFiles` doesn't set `PostId` on the destination.** When a card is copied to another board, the new `FileInfo` is saved without setting `PostId` to the destination board's ID. Those copied files permanently fall into the legacy fallback path and bypass ownership validation, even after the migration.

server/app/files.go
Rajat-Dabade: Agree,  added data migration.

server/app/files.go
Rajat-Dabade: Yes, nice catch. Updated the code, now the `boardID` is included in the path.

Rajat-Dabade:
> Running it through Claude it also found something we might need to address: CopyCardFiles doesn't set PostId on the destination. When a card is copied to another board, the new FileInfo is saved without setting PostId to the destination board's ID. Those copied files permanently fall into the legacy fallback path and bypass ownership validation, even after the migration.

@devinbinnie `CopyCardFiles` already uses `getDestinationFilePath`, the same function that `SaveFile` uses for uploads. Copied files will land at `boards/YYYYMMDD/{boardID}/{filename}` automatically, so now no changes are needed.

server/app/files_test.go
```diff
@@ -207,51 +205,49 @@ func TestGetFileReader(t *testing.T) {
...
 
-		writeFileErrorFunc := func(reader io.Reader, filePath string) error {
+		th.Store.EXPECT().SaveFileInfo(gomock.Any()).DoAndReturn(func(info *mm_model.FileInfo) error {
+			parts := strings.Split(filepath.ToSlash(info.Path), "/")
+			assert.Equal(t, "boards", parts[0])
+			assert.Equal(t, testBoardID, parts[2]) // boardID is the 3rd path component
+			fileName = parts[len(parts)-1]
 			return nil
-		}
+		})
```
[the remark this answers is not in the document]
Rajat-Dabade: This was part of an earlier design where we stored the board ID in PostId. We completely abandoned that approach and switched to path-based ownership (boards/YYYYMMDD/{boardID}/{filename}).

server/services/store/sqlstore/data_migrations.go
```diff
@@ -903,3 +907,165 @@ func (s *SQLStore) runPostgresDeDuplicateCategoryBoardsMigration() error {
...
+//  1. Extracts the filename and the block's boardID.
+//  2. Looks up the FileInfo to get the current storage path.
+//  3. If the path is the old 3-part format, moves the file via fb and updates FileInfo.Path.
+//
+// Files referenced by more than one board are migrated to the board of the first block
+// encountered; subsequent references are skipped with a warning.
+//
+// The migration is idempotent: it records completion in the system-settings table and
+// is a no-op if it has already run successfully.
+func (s *SQLStore) RunFileOwnershipMigration(
```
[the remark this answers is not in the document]
Rajat-Dabade: I have updated the PR objective.

server/app/files_test.go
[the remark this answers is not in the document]
Rajat-Dabade: We can resolve this one.

server/services/store/sqlstore/data_migrations.go
```diff
@@ -903,3 +907,165 @@ func (s *SQLStore) runPostgresDeDuplicateCategoryBoardsMigration() error {
...
+// (boards/YYYYMMDD/filename) to the new 4-part path (boards/YYYYMMDD/boardID/filename)
+// so that board ownership can be verified from the path alone.
+//
+// For each image/attachment block it:
+//  1. Extracts the filename and the block's boardID.
+//  2. Looks up the FileInfo to get the current storage path.
+//  3. If the path is the old 3-part format, moves the file via fb and updates FileInfo.Path.
+//
+// Files referenced by more than one board are migrated to the board of the first block
+// encountered; subsequent references are skipped with a warning.
```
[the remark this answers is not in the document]
Rajat-Dabade: Using a file from one board in another board (cross-board fileId reuse) is not a valid or supported behaviour. In normal usage, this never happens. When you duplicate a board or a card, the system always creates a new file ID using CopyCardFiles. So if we see a file being reused across boards, it’s likely due to an old bug or some unintended behaviour. Because of that, breaking this behaviour is the right thing to do.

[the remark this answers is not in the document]
Rajat-Dabade: We can resolve this one.

server/services/store/sqlstore/data_migrations.go
```diff
@@ -903,3 +907,165 @@ func (s *SQLStore) runPostgresDeDuplicateCategoryBoardsMigration() error {
...
+	// processedFiles tracks files we have already migrated so that a file
+	// referenced by multiple blocks is only moved once.
+	processedFiles := make(map[string]bool)
+
+	for _, blockType := range []model.BlockType{model.TypeImage, model.TypeAttachment} {
+		for page := 0; ; page++ {
+			opts := model.QueryBlocksOptions{
+				BlockType: blockType,
+				Page:      page,
+				PerPage:   fileOwnershipMigrationBatchSize,
```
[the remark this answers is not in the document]
Rajat-Dabade: attachmentId is not used in the upload flow; files are always stored under fileId, no change needed.

server/services/store/sqlstore/data_migrations.go
```diff
@@ -903,3 +907,165 @@ func (s *SQLStore) runPostgresDeDuplicateCategoryBoardsMigration() error {
...
+			if len(blocks) < fileOwnershipMigrationBatchSize {
+				break
+			}
+		}
+	}
+
+	s.logger.Debug("== File ownership migration complete ====================")
+
+	if err := s.SetSystemSetting(FileOwnershipMigrationKey, "true"); err != nil {
+		s.logger.Error("RunFileOwnershipMigration: failed to mark migration complete", mlog.Err(err))
```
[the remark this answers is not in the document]
Rajat-Dabade: Done.

Rajat-Dabade:
@coderabbitai  can you please re-review it.

Rajat-Dabade:
@devinbinnie A friendly reminder to review this PR. Thank you.

server/app/initialize.go
```diff
@@ -10,12 +10,26 @@ import (
...
+		// RunFileOwnershipMigration is a *sqlstore.SQLStore method not in the Store
+		// interface (the generator can't handle cross-package parameter types), so we
+		// call it via a narrow structural type assertion.
+		type fileOwnershipMigrator interface {
+			RunFileOwnershipMigration(
+				moveFile func(oldPath, newPath string) error,
+				fileExists func(path string) (bool, error),
+			)
+		}
+		if m, ok := a.store.(fileOwnershipMigrator); ok {
```
devinbinnie: This seems like it's in the wrong place. Shouldn't this go with the other migrations in `migrate.go:runMigrationSequence()`? https://github.com/mattermost/mattermost-plugin-boards/blob/7d47a1163e616887a921dbed1b19ecea53fa563d/server/services/store/sqlstore/migrate.go#L241?

If so, we can probably get rid of the `skipMigrations` flag too, I'm not sure where that's used.

server/services/store/sqlstore/files.go
```diff
@@ -97,6 +97,25 @@ func (s *SQLStore) saveFileInfo(db sq.BaseRunner, fileInfo *mmModel.FileInfo) er
 	return nil
 }
 
+func (s *SQLStore) updateFileInfoPath(db sq.BaseRunner, fileInfoID, newPath string) error {
```
devinbinnie: Since this is just used by the migration, we probably don't need it to be its own function since it wouldn't be used by the application otherwise.

server/services/store/sqlstore/data_migrations.go
```diff
@@ -903,3 +907,163 @@ func (s *SQLStore) runPostgresDeDuplicateCategoryBoardsMigration() error {
...
+
+	newPath := filepath.ToSlash(filepath.Join("boards", parts[1], boardID, parts[2]))
+
+	exists, err := fileExists(normalizedOldPath)
+	if err != nil {
+		return fmt.Errorf("checking existence of %s: %w", normalizedOldPath, err)
+	}
+
+	if exists {
+		if err := moveFile(normalizedOldPath, newPath); err != nil {
```
devinbinnie: So unfortunately I think there's a problem with this. We don't know how the files may be stored for the server that has this plugin configured. Some servers may be using S3 as a backend, meaning we can't just assume its on a file system that we have access to move files around for.

Unfortunately what that means (and excuse my oversight) is that older files cannot be migrated over to use the new path structure. However I think we can still use the new path structure for new files.

So this means we'll have to account for 3 possible path structures:
- `{teamId}/{boardId}/{filename}` - this one we can keep the same, and we can pull out the `boardId`
- `boards/{date}/{filename}` - this one we'll have to infer the board ID somehow. If it's not too performance heavy we could check this ad-hoc and avoid a data migration altogether
- `boards/{date}/{boardId}/{filename}` - same as the first one

This would mean that:
- no files need to be moved
- we can still get the board ID regardless of which path was used, we'll just have to account for the current method

I would personally prefer to avoid a data migration if we can, but that should only be if the ad-hoc method of getting the board ID is performant enough.

devinbinnie (changes requested):
Unfortunately we'll have to modify the approach a bit here. See the comments I've made below. Happy to discuss whenever you have time if you'd like :)

server/services/store/sqlstore/data_migrations.go
Rajat-Dabade: Okay, if that's the case, then yes, we can drop the data migration, and for the file store at `boards/{date}/{filename}`, we can do the following: 

- First, find the block associated with the file. 
- Find the boardId associated with the block.
- If a block is found and it belongs to the same board as the incoming request, then allow
- If a block is found, but it belongs to a different board, then reject
- If no block is found yet (fresh upload, not yet attached to any block), then allow. 
- And all the new files will be stored in the new format `boards/YYYYMMDD/{boardID}/{filename}`. 

Now that I was looking into the code, we already have the function `validateFileReferencedByBoard` which does the same thing, and it is not performance-heavy because the query Filters by board_id (indexed column) and type (image or attachment), a small subset of all blocks

server/app/initialize.go
Rajat-Dabade: Not required now. Reverted the data migration code.

server/services/store/sqlstore/files.go
Rajat-Dabade: Not required now. Reverted the data migration code.

Rajat-Dabade:
@coderabbitai please re-review this PR.

Rajat-Dabade:
@coderabbitai, we have removed the complete data migration part, so you can skip the data migration issue. Regarding issue 4: Both the function `validateFileReferencedByBoard` and `filterAuthorizedFilesForBoard` serve different purposes. `validateFileReferencedByBoard` validates a new file reference; it only needs active blocks, so `GetBlocksWithType` is correct there. `filterAuthorizedFilesForBoard` is called when restoring a previously deleted block. The fle being restored may have no active blocks referencing it, so `GetBlocksWithType` would find nothing and incorrectly deny the restoring. The performance concern is valid, but the right fix is to add a block-type filter so the query only returns relevant history rows. Switching to active blocks would break the undelete behaviour, so that’s not the right approach. For now, you can skip issue 4 above. 

Please re-review it.

server/app/files.go
```diff
@@ -243,13 +296,14 @@ func getDestinationFilePath(isTemplate bool, teamID, boardID, filename string) (
...
-	// which deletes all files inside the "date" subdirectory
+	// Template files are stored at teamID/boardID/filename (no "boards/" prefix)
+	// so they can be identified by path alone and are not affected by DataRetention.
 	if isTemplate {
 		return filepath.Join(teamID, boardID, filename), nil
 	}
-	return filepath.Join(utils.GetBaseFilePath(), filename), nil
+	// Regular files: boards/YYYYMMDD/{boardID}/{filename}
+	// The boardID in the path makes ownership verifiable without extra DB lookups.
+	return filepath.Join(utils.GetBaseFilePath(), boardID, filename), nil
```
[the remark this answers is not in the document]
Rajat-Dabade: The fallback in `GetFilePath` only runs when fileInfo is missing or has no path. In practice, every file already has a FileInfo record with the correct `boards/YYYYMMDD/{boardID}/filename` path set during upload, so the function just reads from `fileInfo.Path` and never hits the fallback.

Changing the fallback to use `GetBaseFilePath()` would actually be incorrect, since it uses today’s date and would fail to find older files. So no change is needed here.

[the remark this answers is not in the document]
Rajat-Dabade: We can remove it right?

server/services/store/sqlstore/data_migrations.go
```diff
@@ -903,3 +907,163 @@ func (s *SQLStore) runPostgresDeDuplicateCategoryBoardsMigration() error {
...
+		return nil // filename too short to have a type prefix; skip
+	}
+	fileInfoID := base[1:]
+
+	fileInfo, err := s.GetFileInfo(fileInfoID)
+	if err != nil {
+		return err
+	}
+	if fileInfo == nil {
+		return nil
```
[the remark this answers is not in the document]
Rajat-Dabade: We are not using data migration anymore, you can resolve this one.

server/app/blocks.go
```diff
@@ -80,6 +80,20 @@ func (a *App) PatchBlockAndNotify(blockID string, blockPatch *model.BlockPatch,
 		return nil, err
 	}
 
+	// Reject patches that reference files not belonging to this board.
+	if blockPatch.UpdatedFields != nil {
+		if fileID, ok := blockPatch.UpdatedFields[model.BlockFieldFileId].(string); ok && fileID != "" {
+			if ownershipErr := a.validateFileOwnershipForBlockWrite(board.TeamID, oldBlock.BoardID, fileID); ownershipErr != nil {
```
devinbinnie: This logic of pulling out the block by an id is repeated a lot, we could extract to a helper.

server/app/files.go
```diff
@@ -93,18 +93,83 @@ func (a *App) ValidateFileOwnership(teamID, boardID, filename string) error {
 	if err != nil {
 		return err
 	}
-	if fileInfo != nil && fileInfo.Path != "" && fileInfo.Path != emptyString {
-		expectedPath := filepath.Join(teamID, boardID, filename)
-		if fileInfo.Path == expectedPath {
-			return nil
+	if fileInfo != nil {
+		normalizedPath := filepath.ToSlash(fileInfo.Path)
+		if !strings.HasPrefix(normalizedPath, "boards/") {
```
devinbinnie: Looks like this is also repeated below.

devinbinnie (changes requested):
Approach looks better :)

Just a few comments on the code now.

server/app/files.go
Rajat-Dabade: Done.

server/app/blocks.go
Rajat-Dabade: Done.

devinbinnie (approved):
Thanks @Rajat-Dabade!

Rajat-Dabade:
@esarafianou I friendly remainder to review this PR. Thank you.

yasserfaraazkhan:
@Rajat-Dabade the Artifact on the PR is not working. I'll try these locally. But its better if you can inform SRE team to fix this or check if any code change has caused it.

Rajat-Dabade:
@yasserfaraazkhan, yes. But what I can do is I can build it locally and upload it to the Spinwick server just to unblock you.

Rajat-Dabade:
@yasserfaraazkhan, I have uploaded the artifacts build locally on this branch. Can you please test it?

yasserfaraazkhan (approved):
thank you Rajat

server/app/files.go
```diff
@@ -123,24 +195,16 @@ func (a *App) validateFileReferencedByBoard(boardID, filename string) error {
 		return err
 	}
 
-	// Check image blocks
 	for _, block := range imageBlocks {
 		if fileID, ok := block.Fields[model.BlockFieldFileId].(string); ok && fileID == filename {
 			return nil
 		}
-		if attachmentID, ok := block.Fields[model.BlockFieldAttachmentId].(string); ok && attachmentID == filename {
```
edgarbellot, line 131: I see the `attachmentId` field check was removed from the block scan loops here at [validateFileReferencedByBoard](https://github.com/mattermost/mattermost-plugin-boards/blob/420c388188789022d986da333dc158faa103a185/server/app/files.go#L198-L208) (198-208). Previously both `fileId` and `attachmentId` were checked for image and attachment blocks.

This scan acts as the fallback for legacy files (old 3-part paths with no `boardID`). If there are any legacy attachment blocks where the file reference is stored only in the `attachmentId` field and not in `fileId`, this change could cause legitimate ownership lookups to fail - resulting in false access denials for existing users.

Could there be legacy blocks in production that only have `attachmentId` set? If so, would it be safer to keep checking both fields here, similar to how [validateFileRefsInFields](https://github.com/mattermost/mattermost-plugin-boards/blob/420c388188789022d986da333dc158faa103a185/server/app/blocks.go#L188-L200) (188-200) in `blocks.go` already checks both?

server/app/blocks.go
```diff
@@ -80,6 +80,14 @@ func (a *App) PatchBlockAndNotify(blockID string, blockPatch *model.BlockPatch,
...
 
+	// Reject patches that reference files not belonging to this board.
+	if blockPatch.UpdatedFields != nil {
+		err = a.validateFileRefsInFields(board.TeamID, oldBlock.BoardID, blockID, blockPatch.UpdatedFields)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	err = a.store.PatchBlock(blockID, blockPatch, modifiedByID)
```
edgarbellot, line 91: This isn't something introduced by this PR. Adding this comment in this random line because this affects other file.

The new `validateFileRefsInFields` check is applied in `InsertBlockAndNotify`, `InsertBlocksAndNotify`, `PatchBlockAndNotify`, and `PatchBlocksAndNotify` - however there are two additional block write paths in this file that go directly to the store without any file ownership validation:

- [CreateBoardsAndBlocks](https://github.com/mattermost/mattermost-plugin-boards/blob/420c388188789022d986da333dc158faa103a185/server/app/boards_and_blocks.go#L13-L59) (13-59) - used by the archive import flow at [import.go](https://github.com/mattermost/mattermost-plugin-boards/blob/420c388188789022d986da333dc158faa103a185/server/app/import.go#L271) (271)
- [PatchBoardsAndBlocks](https://github.com/mattermost/mattermost-plugin-boards/blob/420c388188789022d986da333dc158faa103a185/server/app/boards_and_blocks.go#L61-L101) (61-101)

Both call `a.store.CreateBoardsAndBlocks` / `a.store.PatchBoardsAndBlocks` directly, bypassing the new validation entirely.

For the import case specifically, `fixImagesAttachments` does remap file IDs afterwards via `PatchBlocks` (which now validates), but a crafted archive could still insert blocks referencing victim file IDs through `CreateBoardsAndBlocks` before that remapping runs. If the remap fails or is skipped (e.g. template boards at [import.go](https://github.com/mattermost/mattermost-plugin-boards/blob/420c388188789022d986da333dc158faa103a185/server/app/import.go#L118-L120) (118-120)), those references would persist.

Would it make sense to add the same `validateFileRefsInFields` pattern to these two functions as well, to ensure all block write paths are covered?

edgarbellot (changes requested):
@Rajat-Dabade great work! I left some comments - let me know if you have any questions :)

server/app/files.go
Rajat-Dabade: @edgarbellot,  the 	`attachmentId` was not used in the first place to store a file, and the check was redundant. For storing any kind of attachment, we always use `fileId`, and based on the type property, we decide which section the file should go to, either the attachment section or the card description section. We have to remove the check from the other part of the code, which I will do in a subsequent PR. But yes, this check is not required.

server/app/blocks.go
Rajat-Dabade: Good catch @edgarbellot,  both cases are handled now.

edgarbellot (approved):
@Rajat-Dabade LGTM, thank you!

PR #181, opened 2026-03-24
Delete attachment redirect to the error page

Rajat-Dabade:
#### Summary
Deleting an attachment could succeed on the server while the UI still showed it, the confirm dialog stayed open, or the app crashed and sent users to the error page. This PR fixes client handling of attachment deletes and websocket tombstones.

The PR includes: 
- Await async `onDelete` and close the confirmation dialog after a successful delete.
- Harden `getUploadPercent` / upload-progress updates when the attachment row is already removed from Redux.
- Fix `updateAttachments` reducer (no premature return that skipped batch items, safer parentId for tombstones).
- Route websocket batches by block type; include attachment tombstones whose hydrated type is wrong but the id is still a known attachment (selectBlocksForAttachmentSliceUpdate).
- Optimistic removal in mutator.deleteBlock after HTTP success for attachment blocks.
- hydrateBlock: unknown/empty type falls back to createBlock without assertFailure.
- updateAllBoardsAndBlocks: pass all boards to updateBoards (not only deleted); align attachment/content filtering with the board page websocket handler.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68059

### DIFF
--- a/server/app/blocks.go
+++ b/server/app/blocks.go
@@ -80,6 +80,14 @@ func (a *App) PatchBlockAndNotify(blockID string, blockPatch *model.BlockPatch, 
 		return nil, err
 	}
 
+	// Reject patches that reference files not belonging to this board.
+	if blockPatch.UpdatedFields != nil {
+		err = a.validateFileRefsInFields(board.TeamID, oldBlock.BoardID, blockID, blockPatch.UpdatedFields)
+		if err != nil {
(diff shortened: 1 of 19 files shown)


Rajat-Dabade:
Done as a part of this PR: https://github.com/mattermost/mattermost-plugin-boards/pull/184

PR #198, opened 2026-05-23
MM-68841: Strip SchemeAdmin when adding guest board members

jgheithcock:
## Summary

- Strip `SchemeAdmin` for guest users in `AddMemberToBoard` before persisting membership, matching the existing guest check in member updates.
- Covers board archive import and any other callers that add members through the app layer.

## Ticket

https://mattermost.atlassian.net/browse/MM-68841

## Release Note

- Fixed an issue where guest users could be assigned board admin when added to a board through certain membership flows.

## Test plan

- [x] Unit test: `TestAddMemberToBoard/should_strip_SchemeAdmin_for_guest_users`
- [x] Verified the unit test fails when the guest check is reverted
- [x] Integration test: `TestUpdateMember/should_strip_admin_role_when_adding_a_guest_member_with_SchemeAdmin_true`

## Change Impact: 🟡 Medium

**Regression Risk:** Localized to board membership persistence (AddMemberToBoard) and user helper (UserIsGuest). The change enforces stripping SchemeAdmin for guest users and adds a system-user short-circuit for guest checks. Unit and integration tests cover membership flows and archive import, reducing risk, but callers outside the exercised paths (or uncommon import/template flows) could previously have persisted admin roles for guests and may be affected. The UserIsGuest short-circuit is a behavioral tweak to a helper used across membership logic.

**QA Recommendation:** Moderate manual QA: validate board member add/update flows (including app-layer callers), archive export/import with modified archives, and template/system-user import paths to ensure guest SchemeAdmin is consistently stripped while non-guest admin assignments remain unchanged.

*Generated by CodeRabbitAI*


server/app/boards.go
```diff
@@ -531,6 +531,14 @@ func (a *App) AddMemberToBoard(member *model.BoardMember) (*model.BoardMember, e
 		return existingMembership, nil
 	}
 
+	isGuest, err := a.UserIsGuest(member.UserID)
```
nang2049, line 534: This fix correctly closes new additions but the early return above means any guest previously saved with SchemeAdmin=true via handleAddMember or archive import keeps it. a small SQL migration demoting existing guest admin rows should ship with the PR. If it's theoretical a note in the release note is probably enough. Or maybe worth checking with security team?

server/integrationtests/board_test.go
```diff
@@ -1855,6 +1855,42 @@ func TestUpdateMember(t *testing.T) {
 		require.True(t, updatedGuestMember.SchemeEditor)
 		require.False(t, updatedGuestMember.SchemeAdmin)
 	})
+
+	t.Run("should strip admin role when adding a guest member with SchemeAdmin true", func(t *testing.T) {
```
nang2049, line 1859: Nice to have one more test that runs an archive import containing a guest member with SchemeAdmin: true and verifies the stored role.

nang2049 (commented):
Thanks @jgheithcock I added some suggestions!

server/app/boards.go
jgheithcock: I'm inclined to keep this ticket focused on just the reported issue. I'd also suggest an alternative is to have some admin script a user could run for cleanup if this is an issue. That said, @hmohammed-prodsec, what is your take?

server/integrationtests/board_test.go
jgheithcock: Good point, added in b4ccfdb

jgheithcock:
@coderabbitai resolve

hmohammed-prodsec (approved):

PR #212, opened 2026-06-18
MM-69067: Revoke Board Admin from users demoted to System Guest

nang2049:
#### Summary
Aligns Boards with the channel behavior on guest demotion. Previously, when a member who held Board Admin on one or more boards was changed to System Guest the per-board SchemeAdmin flag was not reconciled with the new system role so the user's effective role on those boards did not match their new system role.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69067

#### QA
Needs one System Admin and one regular team Member on the same team.
1. As System Admin create a private board. Open Share - add the Member as Admin.
2. As the Member (second browser) open the board. Confirm admin-only UI is visible: Share dialog can change roles and board type can be changed.
3. As System Admin go to System Console - User Management - Users find the member, and change their role to Guest.
4. As the demoted user, reload Boards and reopen the same board.
Post-fix: admin-only controls are gone, the user only sees the role they actually have and any admin-gated request returns 403.

## Change Impact: 🔴 High

**Regression Risk:** Changes harden board authorization by sanitizing persisted `SchemeAdmin` when the user is (or cannot be verified as non-) a System Guest, and they add a login-time revocation path. This touches the core permission evaluation flow (hot authorization path) and extends the permissions store contract with `GetUserByID`, increasing the chance of behavioral regressions (e.g., legitimate admin users being denied if user lookups fail or are stale). The logic is security-defensive (fails closed for lookup errors), which reduces privilege-escalation risk but can still cause incorrect 403 denial if integration behavior differs from assumptions.

**QA Recommendation:** Perform focused manual security/UX verification on top of automated coverage: (1) Demote a Board Admin to System Guest and confirm all board-admin UI controls disappear (including Share dialog and board type change) immediately after login; (2) Verify admin-gated API actions (board type change, relinking channels, membership modification, board deletion) return 403 for demoted guests; (3) Verify non-demoted Board Admins retain access; (4) Confirm no transient access remains across concurrent requests during/around demotion.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/app/user.go
+++ b/server/app/user.go
@@ -6,6 +6,7 @@ package app
 import (
 	"github.com/mattermost/mattermost-plugin-boards/server/model"
 	mmModel "github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
 )
 
 func (a *App) GetTeamUsers(teamID string, asGuestID string) ([]*model.User, error) {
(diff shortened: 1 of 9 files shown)


avasconcelos114 (approved):
LGTM!

edgarbellot (approved):
LGTM, thank you!!

PR #213, opened 2026-06-18
MM-69280: Filter undefined blocks before sorting updated time/by properties

jgheithcock:
#### Summary

Boards could crash when rendering the "Last updated time" or "Last updated by" properties on cards with no content blocks and no comments. `getLastCardContent` and `getLastCardComment` can return `undefined`, but both property components passed those values into a sort comparator that accessed `.updateAt`. This is unsafe in general and surfaced as a page crash under Chrome 149's V8 sort regression (crbug 502534423) on boards with many cards.

This change filters out undefined blocks before sorting and removes the incorrect `as Block` casts. The unit test targets that underlying condition (undefined selector results) rather than reproducing the ticket's ~500-card Chrome 149 scenario, as Jest's V8 doesn't crash when `Array.sort` encounters undefined blocks.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-69280

#### Release Note

Fixed an issue where Boards could crash when displaying the Last updated time or Last updated by property on cards with no content or comments.

#### QA

1. Create a board with the "Last updated time" property visible.
2. Add a card with no content blocks and no comments.
3. Confirm the board renders normally and the property shows the card's update time.

Optional (Chrome 149): reproduce with ~500 cards per the ticket; confirm the board no longer crashes.

## Change Impact: 🟡 Medium

**Regression Risk:** Medium—sorting/selection logic is centralized into a new shared helper used by both “Last updated time” and “Last updated by”. While the change is localized to rendering and adds `undefined` filtering to prevent the original crash, the shared utility introduces a single dependency point and relies on consistent `updateAt` values across candidates.

**QA Recommendation:** Low-to-Medium manual QA. Verify the original crash scenario (card with no content blocks and no comments while the property is visible), plus one mixed board case (some cards with comments/content, some without) and confirm both properties still pick the newest block correctly.
*Generated by CodeRabbitAI*

### DIFF
--- a/webapp/src/properties/latestUpdatedBlock.test.ts
+++ b/webapp/src/properties/latestUpdatedBlock.test.ts
@@ -1,0 +1,26 @@
+// Copyright (c) 2020-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {createCard} from '../blocks/card'
+import {createCommentBlock} from '../blocks/commentBlock'
+
+import {definedBlocksForLatestUpdate, getLatestUpdatedBlock} from './latestUpdatedBlock'
+
+describe('properties/latestUpdatedBlock', () => {
+    test('excludes undefined content and comment blocks before sorting', () => {
+        const card = createCard()
+        card.updateAt = Date.parse('10 Jun 2021 16:22:00')
+
+        expect(definedBlocksForLatestUpdate(card, undefined, undefined)).toEqual([card])
+    })
+
+    test('returns the block with the most recent updateAt', () => {
+        const card = createCard()
+        card.updateAt = Date.parse('10 Jun 2021 16:22:00')
+
+        const comment = createCommentBlock()
+        comment.updateAt = Date.parse('15 Jun 2021 16:22:00')
+
+        expect(getLatestUpdatedBlock(card, undefined, comment)).toBe(comment)
+    })
+})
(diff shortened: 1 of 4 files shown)


webapp/src/properties/latestUpdatedBlock.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    test('returns the block with the most recent updateAt', () => {
+        const card = createCard()
+        card.updateAt = Date.parse('10 Jun 2021 16:22:00')
+
+        const comment = createCommentBlock()
+        comment.updateAt = Date.parse('15 Jun 2021 16:22:00')
+
+        expect(getLatestUpdatedBlock(card, undefined, comment)).toBe(comment)
+    })
```
marianunez, line 32: We should add a test for getLatestUpdatedBlock with nil comment and Content to cover for the bug reported

marianunez (approved):

webapp/src/properties/latestUpdatedBlock.test.ts
jgheithcock: Good callout. Added in [3c3a287](https://github.com/mattermost/mattermost-plugin-boards/pull/213/commits/3c3a2879476f3b283d596c68a9f3f2e85bc4bf03)

ogi-m (approved):
Let's regression test post-merge.

PR #216, opened 2026-06-30
MM-69521: prevent Boards notification worker panic from malformed block properties

nang2049:
#### Summary
Hardens the notification subscription worker against malformed block payloads that could panic the worker goroutine and crash the Boards plugin process.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69521

## Change Impact: Medium 🟡

**Regression Risk:** Changes are limited to Boards notification subscription handling and block property validation, but they alter runtime behavior in a background worker and expand validation strictness around `fields.properties`. The panic-recovery path is now covered by new unit tests, reducing risk of unhandled crashes, though malformed payload edge cases can still surface differences in behavior.

**QA Recommendation:** Light manual QA—run a short smoke test for Boards notifications with both well-formed and malformed block payloads, and confirm the worker remains healthy after induced callback panics.  
*Generated by CodeRabbitAI*

### DIFF
--- a/server/model/block.go
+++ b/server/model/block.go
@@ -194,7 +194,7 @@ func (b *Block) baseValidations() error {
 		}
 	}
 
-	if propsIface, ok := b.Fields["properties"]; ok && propsIface != nil {
+	if propsIface, present := b.Fields["properties"]; present {
 		if _, ok := propsIface.(map[string]interface{}); !ok {
 			return ErrBlockPropertiesInvalidType
 		}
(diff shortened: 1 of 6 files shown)


avasconcelos114 (approved):
LGTM!

server/model/block.go
```diff
@@ -193,6 +194,12 @@ func (b *Block) baseValidations() error {
 		}
 	}
 
+	if propsIface, ok := b.Fields["properties"]; ok && propsIface != nil {
```
edgarbellot: nit: `"properties": null` bypasses the type check in `baseValidations()` because the guard condition `ok && propsIface != nil` short-circuits before the type assertion.

edgarbellot (approved):
Looking good 🚀 thank you!

PR #218, opened 2026-07-01
MM-69607: enforce MaxFileSize on archive import

nang2049:
#### Summary
MM-69607 Boards archive import bypasses MaxFileSize. Any authenticated user could write files of any size to the storage backend by importing a crafted archive exhausting disk and taking the whole instance down.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69607

## Change Impact: 🟡 Medium

**Regression Risk:** The fix spans archive import handling and shared file-saving logic, so it affects a user-facing upload path and related error handling. The new size enforcement changes write behavior and cleanup paths, which could surface regressions around oversized files or partial write failures, but the impact is mostly contained and rollback should be straightforward.

**QA Recommendation:** Moderate QA is recommended. Prioritize validating archive imports at, below, and above `MaxFileSize`, plus confirming normal file saves still work and oversized uploads return the expected error.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/api/archive.go
+++ b/server/api/archive.go
@@ -6,6 +6,7 @@ package api
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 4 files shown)


avasconcelos114 (approved):
Great stuff!

PR #224, opened 2026-07-09
MM-69622 Refactoring board validators

avasconcelos114:
#### Summary

This PR refactors some of the checks done on create boards to make it easy to run the same ones in other flows

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69622

## Change Impact: High 🟥

**Regression Risk:** Authorization logic is now centralized and enforced across multiple user-facing board creation paths, including archive import. Archive import adds a new validation pre-pass that requires reader seekability and can abort earlier; this changes failure modes and import control flow (potentially affecting parsing, ordering, and “no partial persistence” behavior).

**QA Recommendation:** Run focused end-to-end permission allow/deny tests across: (1) single board creation, (2) create boards+blocks, and (3) archive import (including multi-board archives and verification that denied boards don’t partially persist). Additionally, manually validate archive import behavior for non-seekable/non-standard inputs to confirm the expected early failure.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/app/import.go
+++ b/server/app/import.go
@@ -293,7 +293,7 @@ func (a *App) parseBoardsAndBlocks(r io.Reader, opt model.ImportArchiveOptions) 
 					block.UpdateAt = now
 					board, err := a.blockToBoard(block, opt)
 					if err != nil {
-						return nil, nil, fmt.Errorf("cannot convert archive line %d to block: %w", lineNum, err)
+						return nil, nil, fmt.Errorf("cannot convert archive line %d to board: %w", lineNum, err)
 					}
 					if err := board.IsValidForImport(); err != nil {
 						return nil, nil, err
--- a/server/app/import.go
+++ b/server/app/import.go
@@ -110,6 +110,20 @@ func (a *App) ImportArchive(r io.Reader, opt model.ImportArchiveOptions) error {
 	}
 }
 
+// validateImportedBoards runs the optional authorization callback against every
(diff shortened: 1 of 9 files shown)


server/app/import.go
```diff
@@ -110,6 +110,20 @@ func (a *App) ImportArchive(r io.Reader, opt model.ImportArchiveOptions) error {
 	}
 }
 
+// validateImportedBoards runs the optional authorization callback against every
+// board parsed from an archive, aborting the import if any board is rejected.
```
nang2049, line 193: validateImportedBoards is invoked from ImportBoardJSONL which ImportArchive calls once per board.jsonl entry and persists immediately via CreateBoardsAndBlocks before the next entry is read. For multi-board archives if board #2 fails auth, board #1 is already committed.

Suggestion: hoist validation into ImportArchive, drop the per-jsonl call afterwards.

nang2049 (commented):
Thanks @avasconcelos114 just one suggestion!

edgarbellot (changes requested):
@avasconcelos114 I noticed this while verifying the changes - `POST /boards/{boardID}/duplicate` also needs the same check.

The other endpoints are covered (both happy path and edge cases) :)

edgarbellot (approved):
All good, thank you Andre!!

nang2049 (approved):

PR #243, opened 2026-07-30
MM-69978: strengthen validation of card property data

nang2049:
#### Summary
Card property templates and values were accepted and persisted without checking that they matched the shape the clients expect. This adds a shared validator, used across the board, card and block create and patch paths, so data that does not match the expected type is rejected with a 400 rather than stored. The webapp side coerces card property names and values before rendering them, so any existing record that does not match the expected shape degrades gracefully instead of breaking the view it appears in.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69978

## Change Impact: 🟡 Medium

**Regression Risk:** Shared validation and rendering utilities affect server persistence and webapp display paths. Automated tests cover malformed data, but valid edge cases may regress.

**QA Recommendation:** Manual QA can be skipped due to comprehensive automated coverage. Targeted smoke testing remains optional.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/api/blocks.go
+++ b/server/api/blocks.go
@@ -258,6 +258,11 @@ func (a *API) handlePostBlocks(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if err = model.ValidateBlockProperties(block); err != nil {
+			a.errorResponse(w, r, model.NewErrBadRequest(err.Error()))
+			return
+		}
+
 		if block.CreateAt < 1 {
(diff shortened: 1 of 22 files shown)


edgarbellot (approved):
LGTM, thank you Nevy!

avasconcelos114 (approved):
LGTM! Nicely done with the last commit moving the validation checks onto `PatchBlocksAndNotify` :D

PR #254, opened 2026-09-16
Migrate Boards plugin to React 19

nang2049:
#### Summary
Upgrades the Boards webapp from React 17 to React 19 and aligns its dependency stack with the Mattermost webapp, so the plugin runs against the React 19 host.

Dependency changes:
- react/react-dom 17 → 19.2.8, @types/react(-dom) 19
- react-redux 9, redux 5, @reduxjs/toolkit 2, react-intl 7
- `react-beautiful-dnd` → `@hello-pangea/dnd` (unmaintained, React 17 peer only)
- react-dnd 16, @fullcalendar 6, react-hotkeys-hook 4
- TypeScript 5.6, jest 30, @testing-library/react 16
- `overrides` for `react-is`, `react-dnd-scrolling`, `emoji-mart`, and `react-day-picker`

Code changes:
- `JSX.Element` → `React.JSX.Element`
- Widened `RefObject` types for nullable refs;
- Hotkeys: `cmd+` → `meta+`, `del` → `delete` for react-hotkeys-hook 4
- `ReactDOM.render` → `createRoot` in `main.tsx` and `devmain.tsx`
- Replaced deprecated jest matcher aliases and the `jest.mocked(x, true)` signature. Tests were restructured so DOM queries run after the render commits, which React 19 defers. Snapshots were re-recorded for dnd, FullCalendar, and react-select output changes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-70711

## Change Impact: 🟡 Medium

**Regression Risk:** The React 19 migration affects shared frontend components, dependencies, build configuration, drag-and-drop, WebSocket setup, and server retention handling. The broad scope creates moderate regression risk. Public shared-link rendering was reported separately and is unrelated to this change.

**QA Recommendation:** Perform targeted manual QA for board rendering, editing, drag-and-drop, navigation, hotkeys, and WebSocket reconnect flows. Full manual regression testing is not required.

*Generated by CodeRabbitAI*


webapp/package.json
```diff
@@ -34,36 +34,38 @@
...
 		"immutable": "3.7.6",
 		"lodash": "^4.17.21",
 		"marked": "^4.0.12",
 		"mattermost-redux": "10.8.0",
 		"moment": "^2.29.1",
 		"nanoevents": "^5.1.13",
-		"react": "17.0.2",
-		"react-beautiful-dnd": "^13.1.1",
+		"react": "19.2.8",
 		"react-day-picker": "^7.4.10",
```
avasconcelos114, line 57: Should react-day-picker also be upgraded? The peer dependencies for 7.4.10 only include up to react 17 but not sure if this is necessarily problematic

avasconcelos114 (commented):
Only one thing that might need to be confirmed since I'm not sure if this is necessarily problematic. Maybe we can test it first and keep the version as-is if the feature is working fine in v12

webapp/package.json
nang2049: Good catch! the peer range is stale metadata, and I verified it both ways. 7.4.10 doesn't touch anything React 19 removed: no findDOMNode, no legacy context, no legacy lifecycles, no string refs. It uses defaultProps, but only on class components. Functionally both usage sites are covered by tests that assert on the librarys output.

But since v7 is frozen upstream and will never declare React 19 the overrides entry is permanent while we stay on it. 8.10.2 does declare ^19.0.0, so upgrading would let us drop the override and I woild love to do that but I think not in this PR as its quite large. Also bc v8 is a rewrite and not a bump https://daypicker.dev/v8/upgrading 

That's UI-visible work. Happy to file the follow up for removing the override.

avasconcelos114: Ah if that's the case we can definitely leave that for a future round of dependency upgrades that's smaller in scope, thanks for the confirmation!

avasconcelos114 (approved):

ogi-m:
@nang2049 publicly shared links aren't working, the page doesn't render. Can you check if it's related?

image
image

nang2049:
@avasconcelos114 i pushed a fix for the snapshot failures. `webapp/src/manifest.ts` and `server/manifest.go` are generated from `plugin.json` by `make apply`, but this repo commits them. My local `make dist` regenerated them with the version from `git describe` and I committed that by accident. The snapshots were recorded against the no-tag fallback so `sidebar` and `globalHeader` blew up on the version string. Since other plugins all gitignore them so I ended up deleting both files and confirming `make apply` regenerates them.

ogi-m:
Confirmed public boards view is not related to this, created a bug ticket for it.

ogi-m (approved):

